PR #229, opened 2023-10-25
Create es-MX-tutorial.json file

cmdharo:
I created an es-MX-tutorial.json file by copying over the lines from the English file.  There were only minor edits to the file itself (language and region).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -163,9 +163,9 @@
             }
         },
         "node_modules/@babel/core": {
-            "version": "7.23.2",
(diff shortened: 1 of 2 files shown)


amyjko:
Thanks @cmdharo! There's a `package-lock.json`. I'll let it through this time, because these are innocuous, but in the future, we need to start using `npm ci` instead of `npm update` to pull in dependency changes. That will avoid updating `package-lock.json`.

amyjko (approved):
Looking forward to seeing the translation for this!

PR #236, opened 2023-10-30
Es mx output

lindsym:


amyjko (changes requested):
These edits have a few issues:

* There are locale errors: run `npm run locales` to see them, then fix them.
* Many of the @Mention words were translated; those are identifiers that refer to the keys in the JSON file, and are not to be translated. They result in broken links in documentation. These need to be restored to their previous IDs. `npm run locales` will show them to you, but only if you rebase on `main`, since there are some defects in `verify-locales` that this branch doesn't have.
* There are conflicts that you need to resolve. See the pull request in GitHub for a summary of them. Always choose the most recent edit from main for these; there have been many changes to the locale file since this branch was created.

PR #237, opened 2023-10-30
added gallery proj football

mpolasub:
curated a volume based 2d soccer game under "Motion" in wordplay gallery.


amyjko (changes requested):
Thanks for the example! Your code has several conflicts that you need to resolve before I can test and merge this. In particular, all of the references to `ballObj`, `bolaObject`, etc in the Stage expression are invalid. You only localize declarations, not references to names. Please fix these so that there are no conflicts, and resubmit.

amyjko (approved):
What a cute example! I love how you used the mic to trigger the kick.

PR #248, opened 2023-10-31
Create Snail.wp

selenayu000:


amyjko (changes requested):
What a cute little snail! I wonder if we could make this a bit more complex? It seems a bit simple relative to the tutorial examples. Are there some forms of interactivity you could add to it, that might help show some more complex design patterns? For example, maybe the snail's job is to get something, and maybe the snail gets slower the closer it gets to it, because it's a snail?

amyjko:
@selenayu000 Any thoughts?

selenayu000:
I like the idea of the snail getting slower the closer it gets, I can try my best to attempt

amyjko:
@selenayu000 Send me questions if you get stuck!

amyjko:
@selenayu000, any updates on this example?

amyjko:
Any updates on this example @selenayu000? I haven't see n any replies on this for more than a month.

amyjko:
Deleting this PR and branch, since there's no reply from @selenayu000.

PR #249, opened 2023-10-31
Create Christmas.wp

selenayu000:


amyjko (changes requested):
Fun tree dance! Please resolve the conflict in this one before submitting. (And always check for conflicts before you submit). This one binds to `Tree`, but then doesn't use that name, which generates the warning you're seeing.

PR #250, opened 2023-11-01
Update examples.ts

selenayu000:


amyjko (changes requested):
Is there an example that's supposed to be included in this one? I only see the addition of 'Fruits', but no example file.

amyjko:
@selenayu000, can you reply with an update on this? It seems to be missing the `Fruits.wp` example referenced in the update.

PR #259, opened 2023-11-02
Create christmas.wp

selenayu000:


selenayu000:
The code runs in Wordplay with no errors but I am unsure why the tests on Github are failing

amyjko (commented):
Hi @selenayu000! Two issues with this example:

* You haven't added the file name to `examples.ts`, so it won't appear in the gallery. (See the instructions for how to add examples to the gallery)
* Your code has two missing parentheses, and so it fails tests. Correct those errors and then the test will pass. (You should see two yellow underlines in your code, indicating errors).

PR #260, opened 2023-11-02
Create Christmas.wp

selenayu000:


amyjko (changes requested):
* You haven't added the file name to examples.ts, so it won't appear in the gallery. (See the instructions for how to add examples to the gallery)
* Your code has two missing parentheses, and so it fails tests. Correct those errors and then the test will pass. (You should see two yellow underlines in your code, indicating errors).

selenayu000:
Hey Amy, 

I have corrected my code and all tests have passed. I have added my example to examples.ts!

amyjko:
You did the second bullet, but not the first; it's still not added to a gallery in `examples.ts`. I went ahead and added it for you. Review the commit I made to see what you missed.

amyjko (approved):
I added Christmas to the example galleries, so this request is ready now.

selenayu000:
Thank you. I thought I added it but I may have added it to visualizations on accident instead, but after looking at your code I realized I added it in the wrong section

PR #267, opened 2023-11-05
Update examples.ts

selenayu000:


amyjko:
Closing this pull request. In the future, all commits should be in the same updated pull request, not separate pull requests for each commit. (You can update a pull request with new commits as many times as you like; that's how they're intended to be used).

selenayu000:
Thanks for letting me know! Ill keep that in mind for future pull requests

PR #275, opened 2023-11-08
Cn UI locale

iwangy:
I've tried my best to squash any errors - there does seem to be a strange error occurring in the file that I'm not sure how to fix 
image

in the meantime, this document is basically 'done' - however, we definitely do need to revisit to refine the translations.


amyjko (changes requested):
Nice progress @iwangy! There are two issues to resolve before you can fix this:

* There are conflicts you need to resolve. Follow the instructions on the pull request notifying about the conflicts to resolve them. They're likely caused by your version of `zh-CN.json` having strings that `main` has removed or renamed. I suggest choosing `main`'s version for all of them, since its more up to date than your branch.

* The strange errors you mentioned are not strange, they're correct. The locale file has many references to concepts (using the `@Concept`) syntax, that have been translated. The wiki explicitly says not to translate them; the identifiers in concept links refer to keys in the locale file, and are not words or phrases that should be translated. That's what the errors mean.

iwangy:
I think I've gone in and addressed the errors mentioned above! Please let me know if there is any other feedback.

amyjko (changes requested):
When I run `npm run locales`, I still see several dozen locale errors. Are you sure you fixed and committed them? There are many invalid concepts and names reported.

iwangy:
strange i don't see any on my end - ill rerun and verify

amyjko:
My fault, it looks like when I pulled the branch, it failed because of a network connection, so I didn't have your latest commits. I see the updated version now.

amyjko (approved):
Great job pulling all of this together!

amyjko:
Merged! If you're done with this branch, you can delete it, but I'll leave it there in case you're still coordinating on it.

PR #281, opened 2023-11-13
added gallery game 'Where's Waldough'

mpolasub:
Similar to the game where's waldo. Orange shirt man is hidden amongst various emoji. You have to find him and click on him to win.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "wordplay",
-    "version": "0.9.0",
+    "version": "0.9.11",
     "lockfileVersion": 3,
     "requires": true,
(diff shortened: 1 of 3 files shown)


amyjko (changes requested):
Fun idea! A few notes before this is ready:

* The program needs to not have any conflicts (wordplay's word for defect). Fix all of the order of operations conflicts by wrapping expressions in parentheses to disambiguate evaluation order
* It might be nice if the game had some instructions, so the player knows what to do.
* It would also be nice if there was a way to restart the game after playing, other than the IDE's program reset button (but it's fine if the reset button was the intent).

PR #287, opened 2023-11-15
Added the ASCII Video example

willowsalicaceae:
I added the ASCII Video example to the gallery.


amyjko (approved):
This is so cool! I'll definitely be using this as a performance test case. Let's see how high we can get this frame rate :)

PR #302, opened 2023-11-29
Basketball Star Gallery Example

vTempo:


amyjko (changes requested):
Make sure you're not modifying package-lock in this pull request. This suggests you ran npm update, instead of npm ci at some point, changing the versions of packages instead. You'll want to revert these and submit a clean pull requestion.

vTempo:
Is there an easy way to revert the changes? I tried copying the package-lock file on GitHub and re-pushed but I don't think it worked.

amyjko:
The goal is to not include changes to `package.json` or `package-lock.json` in your pull request. So you need to remove those changes from your commits, because we don't want to include those changes in the main branch. Commits have a notion of "staging", which determines which files will be included in a commit; your commits staged package-lock, and they shouldn't have.

To remove files from a previous commit is a bit of a chore, and not always possible:

https://stackoverflow.com/questions/12481639/remove-file-from-latest-commit

Reverting your package-lock is a different problem. You can do the manual copy and paste approach that you did. The more standard way is detailed here:

https://stackoverflow.com/questions/59652396/change-package-lock-json-package-back-to-original-repo-version

All of the above are the complicated ways to do all of this though. The easier way to do this is to just create a clean branch off of main, copy your changed `examples.ts` and example code, then submit a clean pull request.

amyjko:
Closing this abandoned pull request.

PR #304, opened 2023-11-30
Basketball Star project on new branch

vTempo:


amyjko:
It could be a defect in the test, or in the program analysis engine. I'll investigate and report back.

amyjko:
I think I found the cause (infinite recursion in a conflict analysis caused by a defect). I should be able to fix it tonight.

vTempo:
Thank you very much! Should I open an issue about it so maybe someone in verification can look into it as well?

amyjko:
Nope, I know the cause, I just need time to fix it. (Meetings now until 5, then I have some time).

amyjko:
Found a few minutes to fix it. The dependency analysis bug was masking a few conflicts in your program; you can see the changes I made to it to resolve them.

Issue #176, opened 2023-09-20
Arrow keys to modify numeric text fields


amyjko:
## What's the problem?

There are many text fields in the palette that represent numbers (e.g., `offset` or `place` on `Phrase`). Editing them by typing is a bit slow and error prone (changing `9` to `10` requires deleting `10` then typing `9`).

## What's the idea?

Handle arrow up and down within the text field to increment and decrement numeric values, for quick adjustments. The example above would be achieved by just pressing the down arrow to decrement 10 to 9.

## Who benefits?

Anyone using the palette to tweak a design.

## Design ideas

I think simple keyboard shortcuts on the field for up and down would suffice.

wkxdavid:
@amyjko 
Unless I am misunderstanding the issue, I believe the ability to use arrow key navigability for numeric values is working as its in a arrow key navigability slider. Can I get clarification on this?

amyjko:
@wkxdavid This issue refers to text fields, not sliders. I updated the description with a more concrete example of this.

shiinasugioka:
[Pull request](https://github.com/wordplaydev/wordplay/pull/323)

amyjko:
Hi @shiinasugioka and @wkxdavid! Please address the pull request feedback on this issue (including citing this issue in a commit with the text "Fixed #176" so that they are linked.

shiinasugioka:
Hi Amy! I'm going to remove myself from this issue and leave notes for the next person to take a look at. If anyone working on this issue has any questions, feel free to message me on discord (@ Shiina#3331 )

amyjko:
@shiinasugioka, you're not going to finish this pull request then? @wkxdavid, are you going to finish it? If not, you should unassign yourself from the pull request as well, so everyone knows that you're not going to finish it.

shiinasugioka:
@amyjko Actually, sorry I will finish it. I misread the PR comments and thought I had to redo a lot more than I actually had to. I’m not signed up for Wordplay so I thought this would be a good starter issue for a new dev, but looks like I just commit a bunch of file changes that I shouldn’t have. I can get it fixed by the end of the week and I’ll update my PR!

amyjko:
Thanks @shiinasugioka. Let me know if you have any questions; I'll review your revised PR when it's ready.

PR #323, opened 2023-12-06
#176 Arrow Keys modify numeric text field values in Palette

shiinasugioka:
If the text field takes a numeric input, users can press the `arrow up` or `arrow down` key to in/decrement the values.

To make future additions to keyboard bindings, devs can add more cases to the `switch/case` statement.

Devs may want to consider adding a `shift+ArrowUp` or `RightArrow`/`LeftArrow` keybinding that will allow users to increment in 10s as opposed to incrementing one by one.


amyjko (changes requested):
Nice progress on this! Here are a few things to address:

* The strategy for incrementing and decrementing text is interesting. At first, I was concerned, since this applies this behavior to every single text field in the application, meaning that even the login fields and search fields support this. That seemed like too broad a scope. It also doesn't use the `adjust()` functions defined on `NumberLiteral`, `BooleanLiteral`, and `Translation`. But then I thought about it, and realized that this is okay: it's actually useful functionality to have on all fields, since there may be other fields that are not `BindText` fields where this would be helpful (e.g., other forms we might build). This doesn't support Booleans and text as a result, but I think that's okay.
* When up or down is handled, the text field should not also move to the start and end of the text field. (The handler needs to do an `event.stopPropogation`).
* The removal of `stopPropogation` on the handler prevents the caret from moving. Restore the event modifier on the handler declaration. (See what was there before).
* Your commits should reference issue the issue it closes in the commit message, e.g., "Fixed #[issue number]"; this will automatically close the issue on merge to main, and help create a link between commits and issues.
* This branch has a change to package dependencies and `package-lock.json`. Neither should be included. Remove the `0g` dependency change in `package.json` and revert the change to `package-lock`.
* Remove `firebase-debug.log` from the branch. We do not want to commit
* The file should be formatted using `Prettier`. Make sure that's running in your IDE so that it complies to the project's style requirements.

amyjko:
Hi @shiinasugioka, any updates on this pull request?

amyjko:
@shiinasugioka, I haven't seen an update on this pull request for a month. Can you give us an update on its status?

amyjko:
Sigh. Nothing more annoying than a contributor ghosting. I guess I'll finish this work.

amyjko:
This branch is very far behind main and full of gnarly conflicts. I'm pulling out the final revisions to TextField and commiting them separately in a clean commit.

PR #327, opened 2023-12-18
Fixed #191

### LINKED ISSUE #191: Show progress on moderation page
## What's the problem?

When a moderator is moderating projects, there's no way to know how many more there are.

## What's the idea?

Show some kind of remaining number of projects.

## Who benefits?

Moderators, who may get fatigued, and may benefit from knowing how many are left.

## Design ideas

The paginated query in `/moderate/+page.svelte` should be able to return a length, and then we can render a progress counter.

andrewle7:
Incorporated a moderated posts count (to showcase moderator progress), and a unmoderated post count (to showcase total projects to moderate). Incremented the moderated post when user saves a project with new flags, and does not increment when project is skipped. Unmoderated total posts are returned by calling a .length call on the paginated query results.

### DIFF
--- a/src/locale/Locale.ts
+++ b/src/locale/Locale.ts
@@ -34,7 +34,7 @@ export const SupportedLocales = Array.from(
         'es-MX',
         'zh-CN',
         ...(import.meta.hot ? EventuallySupportedLocales : []),
-    ])
+    ]),
 );
 
 /** One of the supported locales above */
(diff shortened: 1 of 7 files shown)


amyjko (approved):
This looks solid! There is the issue that the unmoderated count will have some strange behavior because of batching; it'll count down to zero, then jump up again when the next batch comes. But I don't see a good way around that, since Firebase doesn't have a great way to see how many documents there are in a big query without actually reading all of the documents, which is expensive. We'll leave this behavior intact for now.

I did take this opportunity to localize and format the progress message. This component isn't fully localized yet (there are a few messages still in English only). I'll open a new issue for this.

Issue #331, opened 2023-12-22
Remove non PII strings from project


amyjko:
## What's the problem?

When personally identifiable information appears in text, we show conflicts that allow creators to mark it as not sensitive. Creators may mark it as not sensitive on accident, or it may change in sensitivity, and so there should be a way to unmark text as not sensitive.

## What's the design idea?

In the sharing dialog, we could show a list of strings marked as not sensitive and include a delete button next to them, allowing the set of strings to be edited.

## Who benefits?

Anyone sensitive about the privacy of their data, and other indirect stakeholders who are, such as teachers.

## Design specification

For Projects, we keep track of a list of phrases that have been marked as non PII by the creator (stored in array called nonPII) through the conflict resolution for PII (see related issue: [#310](https://github.com/wordplaydev/wordplay/issues/310)). So in the sharing dialog, we can create a new section displaying all these phrases with a button next to each of them to unmark the text as non PII and add a `PossiblePII` conflict back to that text. This section will also have a reminder to make sure creators know that unmarking text as non PII means that the project will not be saved online anymore.

<img alt="sharing dialog proposal with personal information section at the bottom" src="https://github.com/wordplaydev/wordplay/assets/77032464/ba4489d6-bddb-4c19-9fa5-e83959bd31d5" />

Additionally, to ensure that creators know where they can unmark text as non-sensitive, we can add a note to the conflict resolution reminding them where they can manage PII in their project. 
<img alt="help and conflicts section with updated message telling creators where they can unmark text as non-sensitive" src="https://github.com/wordplaydev/wordplay/assets/77032464/69b20216-c101-40dc-b0be-2c620c44f2dc" />

lwong121:
Hi @amyjko! Above is my initial design based on what I could find from the issue description, related issue, and code. But I would really appreciate any suggestions or feedback on it especially since you worked on the conflict resolutions for PossiblePII. Thanks!

amyjko:
Thanks @lwong121! I like the simplicity of it, and the label you chose; I think that language is consistent with the rest of the language we've used.

A few things to ponder:

[...]

PR #373, opened 2024-02-06
#331 Remove non PII strings from project

lwong121:
Tasks Completed: 
- Added ability for users to unmark text as non-sensitive through the sharing dialog. Now when a user clicks on the "sensitive" button next to some text they previously marked as nonPII, it is removed from the nonPII list and the sharing dialog and the PossiblePII conflict reappears in the conflicts section. 
- Verified that new changes work for different types of PII and that the nonPII list and sharing dialog are updated as expected through manual testing.
- Added translation placeholders for all new properties in the es-MX, example, and zh-CHI files.


amyjko (approved):
This is fantastic work, congratulations on a successful PR! I owe you a high five in person.

Issue #419, opened 2024-03-12
Chinese Locale: Palette Translation Inconsistencies


ahuynh00:
## Palette 
Go to learn, and on the secondary bottom action bar, select palette.

The Chinese locale is missing translations for **format** and **aura** in the palette options, they show up in English. Instances of this are found in the following sections under Learn:
- The Verse: Program, Unparseable
- Its The Little Things: Values, Text, Prefix, Nothing
- Cast Party: Convert, Decisions
- Scene Change: Time, Key, Pointer, Button, Volume, Random, Placement, Chat, Reaction
- Take the Stage: Phrase, Sequence, Scene, Choice
- Callbacks: Memories, Bind, Block, Functions, Structures, Explain

Other instances of English found in the Chinese locale of palette include:
- **Clockwise** and **Counter** in Ensembles: Lists, Sets, and Mappings
- **Color** in Scene Change: Motion, and Take the Stage: Output
- **Straight** in Denouement: Codependency

**The number of options in the palette also varies between English and Chinese throughout the sections of Learn**. In The Verse: Program, the English palette contains 25 options while the Chinese palette contains 20.

## Environment
Desktop

-   mac OS 14.0
-   Safari 17.0
-  Chrome 122.0.6261.111

lpjjj1222:
@amyjko  Hi Amy, I have a commit that solves this issue. Please let me know if there are any changes I need to make~

amyjko:
Per the contributor guidelines, you need to submit a pull request for all proposed changes. Follow the instructions in the guidelines, and I'll provide feedback in the pull request.

amyjko:
This was resolved by PR #464.

PR #438, opened 2024-04-10
Chinese simplified

lpjjj1222:
Solved issue https://github.com/wordplaydev/wordplay/issues/419

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Dates are in `YYYY-MM-DD` format and versions are in [semantic versioning](http:
 -   Defined foreground of footer to ensure button text is visible.
 -   Reset stage output when locales change.
 -   Signficantly reduced memory usage of edit history
+-   Improved preferred spacing of compound data structures and blocks.
 
 ## 0.9.4 2024-05-04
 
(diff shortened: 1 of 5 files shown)


amyjko (changes requested):
There are several conflicts in these files; resolve them and then this can be merged.

amyjko (changes requested):
* Please run `npm run locales` before submitting localization changes; this branch has several localization formatting errors that need to be resolved.
* Many of the templates with parameters (e.g., the `$1` in `the number is $1`) were removed in the Chinese translation. Restore those parameters so that information isn't lost in the descriptions.

benjaminmeow (commented):
They locale file as been updated with new translations and fixed ones that were out of date!

benjaminmeow (commented):
Updating the zh-CN.json with new transaltions and updating old outdated ones. Notice that there is a dummy tutorial file because we are stil debugging it.

amyjko (changes requested):
This looks almost ready, but there appear to be some syntax errors in the locales. For example, throughout, the syntax `$[...]` is used, as in `$[phrase]`, when it you probably mean `$phrase`. `[]` are used in the locale syntax as a conditional.

There are also other syntax errors. For example, `Source > doc` has a code example that is missing the code example escape characters, which is causing the app to crash when creating a new project. 

Remember, in addition to running `npm run locale` to find errors, you want to actually test these translations too, in the live application, to make sure they don't have defects and appear correctly in the interface. Per the documentation, hover over the JSON property keys to see documentation on where they appear in the interface, or what situation they appear in.

Issue #399, opened 2024-02-21
WCAG 1.4.3 Violation for Dark Mode Browser on Homepage


meepsterleapster:
## Expected behavior

Subtitle text under links are clearly readable in all cases and provides the user with a clear understanding of what each  link does

## Actual behavior

Dark mode contrast for subtitles fails WCAG expected contrast ration of 4:5:1. This makes subtitles hard to see on the homepage, creating an accessibility issue.

##Reproduction Steps for Manual Test

1. Go to 'Settings'
2. Click on 'Themes'
3. Select 'Any Dark Mode Theme'
4. Reload Wordplay Page
5. Scroll Down to any gray colored subtitles

_Steps for automatic reproduction coming soon_

## Screenshots
<img width="372" alt="image" src="https://github.com/wordplaydev/wordplay/assets/155786514/19241f0f-4675-4d1a-a906-4f5a9e5a7ed4">

## Environment

Desktop

-   OS + version: [Windows 22H2]
-   Browser + version [Chrome Version 121.0.6167.185 (Official Build) (64-bit)]

meepsterleapster:
If on a device which allows for chromium extensions, a manual test can also be run with with the follow extension (enable the extension to run on the Wordplay Homepage): https://accessibilityinsights.io/downloads/

rf-uw:
<img width="1440" alt="Screenshot 2024-02-21 at 4 27 03 PM" src="https://github.com/wordplaydev/wordplay/assets/156073685/ac1b5c29-b569-4d0c-b626-19f152e3eef5">

Same issue with headers and subheaders in the tutorial and regular code window - highlighted some examples, but the borders and some code symbols (brackets/parentheses, commas, forward slashes) use the same color so are also non-compliant.


<img width="1440" alt="Screenshot 2024-02-21 at 4 47 28 PM" src="https://github.com/wordplaydev/wordplay/assets/156073685/63a5193f-66cb-4e77-94cf-8c7e0834be97">

I'm unsure if these buttons (dark grey in the selected state, black/transparent in the unselected state) qualify as noncompliant - the background color of the buttons doesn't contrast enough with the site background but the text on the buttons does contrast enough with both the buttons and the background.

alexhuynh0:
I am assigning myself to and commenting on this issue. I am open to other people collaborating with me on this issue.

Plans:
- Run Manual Test
- Locate parts of the implementation that are responsible for font color
[...]

PR #458, opened 2024-05-08
Updating subtitle colors in dark mode

leo3tian:
Issue: #399 
Subtitle colors were too dark and violated WCAG guidelines.

Work I did:
Changed the header color from dark gray to white, using the --wordplay-foreground variable in app.html

What the reviewer may need to consider:
From a design standpoint, the reviewer may need to consider whether the color is visually different enough from the rest of the text. The reviewer may also need to consider how this change impacts the rest of the website, other than just the subtitles on the front page. I've provided screenshots at #399, showing how this change impacts other pages. Generally, I think any switch from the previous gray to white should be good for visibility but it's good to check other parts of the website to see if there are any unwanted changes.

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -116,12 +116,12 @@
                 --color-black: #000000;
                 --color-light-grey: #484848;
                 --color-very-light-grey: #373636;
-                --color-dark-grey: #595256;
+                --color-dark-grey: #a3a3a3;
                 --color-shadow: rgb(32, 32, 32, 0.2);
 
                 --wordplay-foreground: var(--color-white);
                 --wordplay-background: var(--color-black);
-                --wordplay-header: var(--wordplay-foreground);
+                --wordplay-header: var(--color-dark-grey);
                 --wordplay-chrome: var(--color-dark-grey);
                 --wordplay-border-color: var(--color-dark-grey);
                 --wordplay-hover: #463000;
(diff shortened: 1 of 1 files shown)


joonpiter:
Hi @leo3tian are you asking for a design review?

amyjko (changes requested):
This fix definitely resolves the specific issue, but it does so by introducing a different issue: inconsistency between the light and dark modes. This change would mean that headers in light mode are the dark grey color, but headers in dark mode are the foreground color. That breaks consistency between the two light/dark modes.

It seems like the change should be in fixing `--color-dark-grey` in dark mode. Isn't that color problematic _everywhere_ in dark mode, not just headers?

leo3tian:
Right, it would be smarter to apply changes to `--color-dark-grey` in dark mode instead of changing the value for `--wordplay-header`. I've changed `--color-dark-grey` to be brighter, but the problem now is that there's a discrepancy in color values:
image
After changing `--color-dark-grey`, dark grey now appears to be lighter than light grey and very light grey. I propose that we get rid of light grey and very light grey altogether - `--color-dark-grey` seems to be light enough already for all elements that we would color grey.

amyjko:
It does create a misalignment between the names of the colors and their relative lightness, but that is common with dark mode: names don't always preserve their semantics when negated for darkness.

Light grey and very light grey are used for other purposes, not just text (e.g., alternating row colors in tables, widget backgrounds). Removing them would break the visual design of those other elements.

amyjko:
Closing unfinished pull request.

PR #464, opened 2024-05-14
solve issue #419.

lpjjj1222:
By addressing the inconsistence of the structure and the inconsistence of the Chinese translations of the same English words in the JSON file, successfully make the hidden Chinese translation words show up.

The number of options in the palette are equal between English and Chinese Learn sections now.


amyjko (approved):
Yay, nice progress on placeholders!

Issue #358, opened 2024-01-25
Issue with scoring in the basketball game


MichelleGuan1128:
## Expected behavior

Expect to earn a score when the basketball is in the hoop.

## Actual behavior

Provide detailed steps that consistently reproduce the behavior, including any example code. If you can, make the steps as minimal as possible, finding the smallest possible set of steps that still produce the unexpected behavior.

1. Go to 'Galleries'
2. Click on 'Games'
3. Scroll down to 'Basketball Star'

## Screenshots
<img width="1440" alt="Screenshot 2024-01-24 at 5 49 08 PM" src="https://github.com/wordplaydev/wordplay/assets/76934364/54137993-5fe9-4aea-b7bf-1d45526d8d93">

<img width="863" alt="Screenshot 2024-01-24 at 5 50 41 PM" src="https://github.com/wordplaydev/wordplay/assets/76934364/4aa5f85e-0b39-4ecb-a868-e6095f8b9822">

Did not earn a score when the basketball is in the hoop. Sometimes it earns a score even if the basketball is not in the hoop. 

## Environment

What platform, OS, browser, and version are you using? Please be exact with version numbers, as different versions can result in a defect happening or not. Don't skip these; cross-browser issues are a common cause of problems.

Desktop
Device: MacBook Pro Intel Core i5
OS: MacOS 14.0
Browser FireFox
Version 118.0.2

amyjko:
Thanks for reporting @MichelleGuan1128! Want to try to fix it?

jianna-braza:
hello! quick question about this issue: is this something I am fixing within wordplay itself or vscode? Is the goal of this issue to fix the current code within wordplay like what is displayed below?

<img width="744" alt="image" src="https://github.com/wordplaydev/wordplay/assets/91649104/f6595e2e-c289-4153-8182-7ab80d883011">

Thank you!

amyjko:
@jianna-braza That's a good question for @MichelleGuan1128. The issue reported is with the logic of the game, not Wordplay, but it may be caused by something in Wordplay (e.g., limitations of the physics engine being used). Your job is to figure out which it is.

amyjko:
It's the end of Winter! Please provide an update on this issue, including:

* Whether you plan to continue work on it
[...]

PR #468, opened 2024-05-18
Update BasketballStar.wp

nAtHanDang0703:
The was an issue regarding the basketball star game, where the player does not earn a score when the basketball is in the hoop. Furthermore, the player sometimes earns a score even if the basketball is not in the hoop. Through trial and error, I changed the range of the basketXPlace and included the y position of the basketball in consideration of the score, since the ball can sometimes drift away from the basket as it goes down. An issue I ran into was that sometimes, the score wouldn't go up if the ball touches the rim and rolls into the basket.


amyjko (approved):
This revision looks great, nice sleuthing!

Fixed #358.

PR #474, opened 2024-05-21
Chinese-simplified

benjaminmeow:
Updated with feedback given. Tested locally and fixed visible syntax errors.


amyjko (changes requested):
This PR includes a new dependency, `node-fetch`, and a corresponding change to `package-lock.json`. These shouldn't be included; not only are they not needed for the project, they're not needed for this PR. Thanks!

benjaminmeow (commented):
I think I must have added node-fetch when I was trying to debug my npm run dev. I thought I was missing a dependency somehow. Should be fixed now. Thanks!

amyjko (changes requested):
There's still a package-lock.json change in there too. Please remove that, otherwise everyone's npm dependencies will be wrong.

benjaminmeow:
I removed all the node-fetch dependencies that I could find in package-lock.json. Hopefully I did it properly this time?

amyjko (changes requested):
This PR still has revisions to `package-lock.json` and `package.json`. Neither should be changed. Please remove them from the PR and then I will merge.

Additionally, there is a change to `en-US.json` that converts the English into Chinese. Please revert that change.

There is also an incorrectly named tutorial file: it was renamed in this PR to `zh_CN-tutorial.json`, but the tutorial file names use dashes, not underscores. Please revert that change.

Davidxuwuhu:
Is anyone still working on this? I can finish this up.

amyjko:
No one is assigned, so it's all yours. It'd be great to finish this locale!

PR #477, opened 2024-05-28
332 resolve duplicate name

euouae682:
Issue: There is no conflict resolution for the DuplicateName conflict (if a declaration ha a duplicate name).

We implemented a resolution button for the duplicate name conflict based on the design specification. Specifically, we did the following:

- Allow the user to remove one of the duplicate names with a button if the duplicate names are on the same line
- Edited representation of the conflict in NodeTexts.ts and locales to include a resolution message
- Display the resolution message next to the button

Some notes:

- The resolution only appears if the duplicate names are on the same line; the user must manually resolve the conflict if there are duplicate names on separate lines.


amyjko (changes requested):
This looks mostly ready! Three minor things to fix:

* There is a change to `package-lock.json` in your branch. Please remove that from the branch; we manage dependencies centrally, and all new dependencies and updates must be approved centrally, so we can verify that updates don't break anything.
* The placeholder in the `es-MX` local should be `$?`, not `TODO`. That's the syntax we use to indicate that a string is unwritten, and it's important, because it tells the platform to fall back to English.
* The English resolution prompt `There are duplicate names, do you want to remove it?` has a grammar error: `it` is ambiguous, because `names` is plural. It should be `There are duplicate names. Do you want to remove one?`

euouae682:
Thanks for the pointers! I've implemented the changes, and it should be ready to merge now!

amyjko (approved):
This looks ready to me, and works in my testing. Approved. Congratulations!

PR #478, opened 2024-05-29
Add completed Traditional Chinese JSON

shockwave656:
- Added completed Traditional Chinese JSON with translations


shockwave656:
Done, I have made all of those changes!

amyjko (changes requested):
The pull request still includes changes to `package.json` and `package-lock.json`. Those need to be removed from this PR before we can merge it.

shockwave656:
Hi Amy! I know you asked me to remove the package.json and package-lock.json files from the traditional Chinese branch. However, when I did that, it told me the branches can't be automatically merged but now that I have the original package files back without any changes, GitHub tells me the branches can be automatically merged. I'm not sure what to do about this, so could you help me through what to do here?

amyjko:
I'm not quite sure what actions you took to remove the changes from your PR, so I can't tell you what might be happening on your end. Let me start from some basics, so I can gauge where you might be getting stuck.

* `package.json` roughly indicates what libraries the project depends on. We don't want random PRs upgrading to new library versions without us testing them first, and that is why I'm asking you to not change that file. `package-lock.json` is the file that indicates which exact versions the project is currently using; it is updated automatically by `npm` when updating or installing packages. The reason those might have changed are likely do to not following the instructions in the wiki, to use `npm ci`, instead of `npm install` or `npm update`
* When you committed these changes to your branch, they became part of your commit history. The only way to undo those commits is to follow quite complicated, risky procedures. You can see some examples [here](https://stackoverflow.com/questions/43762338/how-to-remove-file-from-git-history). What to use depends highly on the state of your local branch. This is one of the major design and usability flaws of Git. Note that all of this is different from *deleting* these files; that's not what you want to do, because that would mean you're proposing to get rid of these files in the project (which would be a disaster for the project).
* The simplest way to undo the changes you made to these two files is to grab the most recent versions of them from the `main` branch, and place those in your local branch, and commit those changes. That would effectively undo it, without having to modify your commit history using those nightmarish changes. It's a bit of a hack, but that would work, because Git is smart enough to know that the file changes you're proposing would make no changes to the files in `main`, and so they would effectively have no impact, because it would be the same text.

Does that help?

shockwave656:
Yeah, that's exactly what I did. I downloaded the most recent versions of the package and package-lock JSON files from the main branch, placed them in my local branch, and committed those changes, so GitHub tells me the branches are able to merge and I also think we should be good to go.

amyjko:
Alright, thanks for confirming. I've verified that there are no changes, despite GitHub showing changes relative to the branch.

amyjko (approved):

amyjko:
Merged, branch deleted. Congratulations on the successful contribution!

shockwave656:
Great, thank you Amy!

PR #516, opened 2024-07-10
474 chinese simplified

Davidxuwuhu:
# Context

Add Chinese localization for the tutorial.
     Revisioned previous translation fixed some conversations that are redundant and unnatural.

Fixed unmatching parts and updated with the latest commits. 
     Translated the new sections being added to the file.

Ensure the variables are defined in English so it is accessible for users.
     Change function names: e.g "@无(none)" to "@None", "@布尔值(boolean)" to “@Boolean”

-   Related Issue #474
-   Closes #474

## Verification

Tested on Microsoft Edge and Google Chrome. Read through the learn section and ensure every part is translated and makes sense.


amyjko (changes requested):
Great progress on the tutorial!

Please run `npm run verify-locales`, per the localization wiki instructions. You'll see several errors in the tutorial, including broken concept links, emotions, and conflicts in example code. Let me know if you need any help resolving them; follow the documentation in the localization wiki for guidelines on the syntax.

Davidxuwuhu:
Hi I fixed most of the errors already but there is still some that are a bit vague. The syntax looks like
    x Unknown tutorial concept:
    x Unknown tutorial concept:
    x Unknown tutorial concept:
But doesn't state which exact part is wrong. Is there anyway I can find out where these eoors occur? Thank you

amyjko:
@Davidxuwuhu I added some more detail about those warnings in `verify-locales.ts`: it now shows the line of dialog.

There are also conflicts in several example programs that don't look like they were translated correctly. Note: if you used machine translation, if does not work on code.

Davidxuwuhu:
Thanks for adding the test cases! I've fixed all the ones I found and it should be working now.

PR #549, opened 2024-08-08
Commit for fixed branch of German localization project.

minkere:
# Context

New pull request for the Wordplay German localization to try to fix the file change issues of the previous pull request.


amyjko (approved):
Looks like a good start!

PR #552, opened 2024-08-18
378 translate the verse

Davidxuwuhu:
Fixed issue 378, which states that 'the verse' has not been correctly translated into the Chinese Simplify tutorial. Also replaced all missing translations for the word 'verse'. The translation now is similar to poem or poetry. 

Meanwhile, translated the titles and subtitles that were still in English. 

-   Closes # 378

## Verification

Tested the Chinese tutorial on Edge, firefox, and Chrome and it functions smoothly. 
Also ran the localization tests and fixed all critical issues.


amyjko (approved):
Looks good!

PR #593, opened 2024-11-04
Fixed #535

### LINKED ISSUE #535: Differentiate tutorial interface highlights by color

When there are two or more tutorial highlights of an interface element, they are both yellow, making it hard to know which element is being referred to.

### What is your idea?

Use multiple colors to distinguish them. Just make a list of standard UI colors already defined in Wordplay, and work through the list to define distinct highlight colors.

SathvikCookie:
# Context

When there is more than one highlight, it becomes difficult to differentiate which highlight links to what. I integrated a system to count the highlights, and number them when more than one highlight exists, to more easily differentiate them. When there is only one highlight, functionality remains the same. Slightly modified the highlight, to remove the opacity change so it is easier to see the,.

## Related issues

-   Closes #535

## Verification

Went through the tutorial to determine its functionality with more than one highlight, and also with exactly one highlight. The primary issue was with contrast on the highlights, so I checked the text color of the highlight and the background color of the highlight to verify contrast was up to standards.

### DIFF
--- a/src/components/app/TutorialHighlight.svelte
+++ b/src/components/app/TutorialHighlight.svelte
@@ -3,6 +3,7 @@
 
     // A class name to highlight.
     export let id: string | undefined = undefined;
+    export let highlightIndex: number | undefined = undefined;
 
     let bounds: DOMRect | undefined = undefined;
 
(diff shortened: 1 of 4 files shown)


amyjko (approved):
Only minor stylistic quibbles here, nothing worth even mentioning. It looks great!

PR #595, opened 2024-11-14
Fix to account for name conflicts during project translation

ahsoj22:
# Context

Translating the above from Spanish to English creates a new name for hello named hola, which creates a duplicate name conflict. I modified translate.ts to detect when a conflict will occur with a new name, and instead of giving it the same name, append a number to 2 to the end of the name, to prevent the conflict.

-   Related Issue 486

## Checklist

<!-- I believe that this pull request is mostly complete however it's definitely possible to do further testing to account for more edge cases.>


amyjko (changes requested):
Thanks looks like it's pretty close! A few points of feedback:

* Your PR should not include changes to versions of dependencies in `package.json` or `package-lock.json` unless those are required to implement functionality. If they aren't, please undo those changes in your PR, just committing source, without changes to configuration
* Thanks for submitting a test with your functionality update!
* You should remove all `console.log` statements from your PR; we want clean logs, and these look like they are for your debugging purposes.
* I see you removed the translation of of camel cased names. Was that intentional? It looks like you might have done this to get the set of original names to be intact, but that removes functionality from the existing translation approach.

ahsoj22:
@amyjko I fixed the console log statements and the dependency changes in the package and package.lock files (I must've made those changes during the debugging process and forgot about them). The camel case functionality must have been removed on accident, and it doesn't whether the test passes or fails or the bug fix so I've added it back.

amyjko (changes requested):
Looks like there are still `package.json` and `package-lock.json` edits in the PR. Otherwise, this looks good to merge.

ahsoj22:
I just copy pasted the files from package.json and package-lock.json from the up-to-date main fork so I'm not sure why it's saying there's changes, it looks like it's complaining about a difference in bracket spacing but it looks the same to me?

amyjko (approved):
Looks like it's just some weird newline difference. That's negligible. I'll approve. Congratulations!

amyjko:
@ahsoj22 This is basically ready to go but it looks like a few Playwright tests aren't passing. That's probably not your fault — they may be flaky, or execution might be slow — but want to investigate to verify?

amyjko:
@ahsoj22 Merged! Congratulations.

PR #603, opened 2024-11-29
#513: Machine translate tutorial

### LINKED ISSUE #513: Machine translate tutorial
## What's the problem?

We have a feature in the `verify-locales` script to translate unwritten locale strings.

## What's the design idea?

Let's extend this to the tutorial, generating draft translations of the English tutorial. The current UI text translation is in `verify-locales.ts`. Updating the `checkTutorial` function is the work to do.

lpjjj1222:
# Context

This PR extends the machine translate feature in the `verify-locales` script to support tutorials.

## Related issues

-   Related Issue #513
-   Closes #513

## Verification

Tested with es-MX-tutorial.json:

1. The file contained an untranslated $? in the Line array.
2. I tested whether the Google Translate API could successfully translate $? by running the updated verify-locales.ts script.
3. After running the script, I verified that the new es-MX-tutorial.json was generated correctly, with the $? replaced by the translated content.

Note:
1. I noticed that the contributor who added $? had made a mistake: there should have been five $? entries in the Line array, representing five lines of dialogue, instead of just one. So, I corrected the file by replacing the single $? with five $?, then re-ran the verify-locales.ts script. The script processed the corrected Line array and generated the expected output, confirming the functionality works as intended.

2. I discovered that the language and region fields in zh-TW-tutorial.json were incorrect while trying to resolve this issue.


amyjko (changes requested):
This looks pretty great @lpjjj1222, nice work. Two two revision requests and one question:

* Can you add English translations of your Chinese comments and error messages? (You can leave the Chinese there; it'll just help others to debug and enhance this in the future to also have clear English comments).
* There are translations of the tutorial missing for several of our pending locales. Can you include translations for them in this PR? That will also help you verify that it works for other full untranslated tutorials. You can just copy the English tutorial for the other locales.
* How do handle cases where a tutorial changes in its structure? (e.g., it is possible to add extra turns in dialog, and even reorder sections of the tutorial). It looks like your script doesn't account for that possibility, and just assumes an identical structure. One way to account for that is to at least check to see if each section has the same concept name, and the same number of turns. If they don't, it probably shouldn't be translated.

lpjjj1222:
Hi Amy @amyjko .Thank you so much for your suggestion~

I have a question regarding the translation process for missing language versions. To provide translations, I would need a template file like example.json, as it contains the $? symbols marking untranslated content.

If I start from scratch and copy and paste the English tutorial as instructed, it won’t be possible to use machine translation since the $? markers are missing. To make machine translation work, I would have to manually change all the content that needs translation to $?, which seems quite cumbersome.

Could you please advise on how to handle this situation?

amyjko:
The functionality would identify missing tutorial files, take the English structure, and generate translations from them, rather than using an existing template file with "needs translation" markers. I think that's a minor change to your script; it just means identifying missing tutorial files, copying the English one, and translating all markers. 

If that seems like too much for the this PR, just let me know. It seems like the functionality isn't quite done if it only works on partially translated and annotated tutorials files. We should also be able to translate the full English tutorial into other languages.

lpjjj1222:
@amyjko Thank you for clarification =）When I tried to implement the functionality to automatically create tutorial files, I encountered some issues. I've already pushed my current code so that you can take a look. 

When I run npm run locales, the newly created tutorial files seem to trigger errors during the check for compliance. For example, it shows an error like:

“Uh oh, there's a conflict in...
Stage([] background: Color(0% 0 0°)), UnexpectedInputs, UnexpectedInputs”

However, when I review the automatically generated files, they don't seem to differ from the original files (since these parts don't need translation, I didn't translate them either).

amyjko:
Hi @lpjjj1222; a few notes:

* The example you gave is suspicious, because of the `UnexpectedInputs` in the code. That is not a valid name, so that would be causing the conflict/error. The other suspicious thing I noticed is that your translated tutorials, such as German, *do* include translated code. All of the examples in the tutorial text have translated names, but the code parts of the tutorial text should not be translated. You could check whether a turn in the tutorial text has any code by looking for the example delimiter (`/`) and if it's present, not translating it.
* As an aside, note that you have a type error in your script. Be sure to fix that before finalizing your PR.

lpjjj1222:
@amyjko Hi Amy, Thank you for pointing out that there is some suspicious translation in the translated tutorials. 

I am also quite confused because I simpliy call the google translation api the way as the locale files being translated.
And As you can see below, some of the translations meet expectation but some do not.

Screenshot 2024-12-06 at 18 02 17

Screenshot 2024-12-06 at 18 01 38

And I have tried to find the place to tell the translation machine not to translate the word begins with a '@'.. (for both locale and tutorial). But I did not find it..

Could you please give me some help?

amyjko:
Google Translate does not know how to distinguish between natural language and the Wordplay rich text syntax. It doesn't know what `@` means, what `/` means. See this section of the localization guide:

lpjjj1222:
@amyjko Thank you for your clarification. Based on this, I try to use regex to match the code and the concept link. 
Screenshot 2024-12-06 at 22 53 52

However, if I use the method shown above, issues may occur occasionally. As shown in the image below, machine translation sometimes adds a Korean particle after @Evaluate. 
Screenshot 2024-12-06 at 22 54 14
In such cases, the concept link cannot be correctly recognized. What I can think of is to modify the code in the first screenshot by changing ```translation = translation!.replace(`${item}`, originalMatch[index]);```to ```translation = translation!.replace(`${item}`, `${originalMatch[index]} `);``` to add a space. However, this causes strange translation results in other cases, such as @Evaluate ?`(redundant space after the concept link)

amyjko:
I do not recommend trying to repair translations; I recommend avoiding translating lines/turns that have any code or concepts in them at all. Google Translate will mangle them unpredictably. Just leave them untranslated, and leave it to a person to translate manually. After this PR is closed, you could open an issue to automate translation of lines with code or concepts, but it requires much more than regular expressions.

lpjjj1222:
Hi @amyjko , I have resolved the issue of avoiding translating lines/turns that contain any code or concepts and converted the previously mentioned Chinese comments into English. However, despite spending considerable time debugging, I couldn’t resolve the remaining issues.  (╥_╥)

lpjjj1222:
Yes that's correct. Much appreciated! @amyjko

amyjko (approved):

amyjko:
Wrapped up this pull request this morning, to prepare for tutorial translation for languages needed for Adrienne's class.

Issue #585, opened 2024-10-17
Apply end-to-end tests for interactive UI components in Project page


yue7xu:
### What happened?

We need thorough **end-to-end tests** for the icons and other interactive UI components on the project page. These tests should verify that each icon performs its intended function correctly, such as starting/pausing animations, adjusting layout or settings, or any other interactive features. For example, some icons do not perform the intended action. For example, clicking "start" may not initiate the animation correctly or other interactive elements like layout adjustment may not trigger as expected.
<img width="1512" alt="interactive icons in project page" src="https://github.com/user-attachments/assets/77966a3f-859d-4dce-a6b5-64b5366db8a6">

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

macOS

yue7xu:
@amyjko Please tell me more so that I can revise this issue with more details and strategies. And can I be assigned to work on this issue?

amyjko:
Assigned! I recommend enumerating every flow you want to test in this issue as a checklist, and then working on the tests, completing the list. Verify all of the tests pass, then submit a PR to merge them into production. I'm happy to elaborate on any of this.

ahsoj22:
@amyjko can I be assigned to work on this issue?

Jan-seon:
@amyjko Hi! Can I also be assigned to work on this issue?

amyjko:
@ahsoj22 declined to work on this verbally at the meetup.

@Jan-seon, are you collaborating with @yue7xu on this? They are already assigned.

Jan-seon:
Yes! We are working together

amyjko:
Okay, added @Jan-seon.

ahsoj22:
I'd like to be added to this issue @amyjko

amyjko:
Great, added @ahsoj22. Be sure to coordinate with the others assigned.

yue7xu:
sorry, I was curious to click around and accidentally click on unassign me. Can you add me back @amyjko?

amyjko:
Added you back. Remember, since there are multiple assignees on this issue, all of you need to be coordinating on the work you're doing. I didn't see any comments here or a Discord channel for this issue. Are you talking to each other in some other channel? (We try to keep all communication here and in Discord). If you're not talking to each other, you're likely to create overlapping and conflicting work, which is definitely something to avoid.

amyjko:
[...]

PR #616, opened 2024-12-05
#585: Command button end2end tests

ahsoj22:
# Context

This issue is related to end to end testing for UI buttons

## Verification

My changes include testing for command button elements, I gave commands uiid identifiers and verified that they rendered on the page when the page is loaded, making sure that the playwright tests passed (if the button actually exists). 

## Checklist
Related to issue #585

This pull request doesn't finish the issue as there are more buttons to go however it's a good starting point for anyone else wanting to work on this issue and needing a simple test to reference from. My next steps are to verify button functionality and figure out a method to create unique identifiers for buttons that appear more than once on the page.

### DIFF
--- a/src/components/editor/util/Commands.ts
+++ b/src/components/editor/util/Commands.ts
@@ -190,6 +190,7 @@ function handleInsert(context: CommandContext, symbol: string) {
 }
 
 export const ShowKeyboardHelp: Command = {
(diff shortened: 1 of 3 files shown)


ahsoj22:
@amyjko

amyjko (approved):

Issue #335, opened 2023-12-23
Resolve unknown name


amyjko:
## What's the problem?

When a name is unknown, but just a typo, it is often hard to notice why the name is unknown

`
number: 1
numbe + number
`

## What's the design idea?

Add a conflict resolution for `UnknownName`, offering to change the name to names that have a Levenstein distance of 1.

## Who benefits?

Anyone looking for a quick fix to the problem.

## Design specification

TBD

amyjko:
It's the end of Winter! Please provide an update on this issue, including:

* Whether you plan to continue work on it
* If you do not, 1) provide a detailed update of its status, 2) add any unfinished work, 3) list any any unmerged branches, and 4) unassign yourself, so someone else can take over the task. Please leave the issue in better shape than you found it, helping future contributors to start from where you left off.

If you do plan to continue work on it, carry on. Otherwise, thank you for your contributions!

amyjko:
No reply :( Unassigning @starter821 @eunjiii-2029344 @choKaylee.

tsutari:
Hey @amyjko I'd like to request to be assigned to this issue.

siddaay:
Hey @amyjko I'd like to request to be assigned to this issue

amyjko:
@tsutari @siddaay Both of you asked for this so I've assigned both of you. You should collaborate on this work. If you're not willing to collaborate, figure out who's taking it and let me now.

A few notes about this:

* This is partially implemented already, but not using Levenstein distance. See what `UnknownName` does; it pulls all names in scope and recommends them, which isn't very precise.
* The design work to do here is minimal. It's mostly to decide things like 1) should the distance be 1 or also 2, to account for multiple typos?, 2) how should they be sorted?, 3) what should be recommended if nothing matches?

siddaay:
Hi @amyjko, we felt good about our current implementation but realized our commits have been failing the "Deploy to Firebase Hosting on merge" test. We're a bit stumped on this as it doesn't seem related to the file we worked on, and were hoping for some guidance.
Here's a screenshot: 
![image](https://github.com/user-attachments/assets/39fc7dce-5180-43ff-9187-c87ccf9cb6d6)
 And our fork, if that may be helpful:
 https://github.com/siddaay/wordplay

amyjko:
[...]

PR #623, opened 2024-12-13
#335 Resolve Unknown Name

siddaay:
# Context

When a name is unknown, but just a typo, it is often hard to notice why the name is unknown. The IDE simply recommends all existing names in the order they were declared. This change in UnknownName implements a Levenshtein distance function and iterates through the list of references returned by getPossibleReferences, eliminating references of Levenshtein distance > 1. We also cap the number of references to check at 50, for the sake of performance. This simply eliminates any names after the first 50 declared from consideration.

Levenshtein distance is a metric quantifying the difference between two strings. We imitated [another implementation](https://gist.github.com/keesey/e09d0af833476385b9ee13b6d26a2b84) which the author made freely available. The only change is that we convert both input strings to lowercase to compare case-insensitively. We felt this made sense as a design decision since forgetting to capitalize could be a common typo. The result is that, for example, if `NewName` is declared and the user inputs `newname`, our implementation will recognize it a distance of 0 and include this in its recommendations. Without case-insensitivity it would be a distance of 2 and therefore ignored.

## Related issues
-   Closes Issue #335 

## Verification
We tested the performance of the feature in Google Chrome using CPU throttling to emulate an older machine. We found a slowdown when testing with large numbers of names (50-100) but this was not significantly impacted by our added functionality, as we tested with and without our code. We timed the additional backend work to take in the range of 0-6ms when capping at 50 references.

## Checklist
This pull request does effectively resolve the issue, but some potential future design changes that could be made if desired:
- [ ] Filter to names within Levenshtein distance 2 instead of 1 and sorting them
- [ ] Make some additional recommendation if no names exist within the desired Levenshtein distance (beyond the default message "I don't know anyone named ... ")
- [ ] Altering the Levenshtein distance function to be case sensitive (delete lines 22-24 in src/conflicts/UnknownName.ts)
- [ ] Change the number of names checked (assign a new value on line 72 in src/conflicts/UnknownName.ts)


amyjko (approved):
This PR is great; the additions are clear, correct, and follow the patterns of the project. The description is complete, well justified. Excellent work!

PR #630, opened 2024-12-26
Update on Issue 606

awang040403:
# Context

Show username for all project previews.

## Related issues

None

## Verification

I set up the environment on my end and tested my edits. I am a little concerned with my work and would like to politely request some help. I am using uid to find id with Creator, while it seems that it cannot find the uid. Did I find id in the wrong place or is it because of limited permission? I have also written the codes for collaborated authors, while I commented it out for now because I am still having issue with single author. This is the path that I downloaded the codes: wordplay\src\components\project. Did I get the issue because of await Creators.getCreator(owner), which might require a longer time to get the username? I will continue to figure it out and test myself. I understand that you are busy, while could you please also see if my codes work on your end if possible? Thanks a lot for helping me with my concerns!

## Checklist

This is a draft pull request for help.

### DIFF
--- a/src/components/app/ProjectPreview.svelte
+++ b/src/components/app/ProjectPreview.svelte
@@ -1,57 +1,40 @@
-<!-- @migration task: review uses of `navigating` -->
 <script lang="ts">
     import type Project from '@models/Project';
     import Evaluator from '@runtime/Evaluator';
-    import { Chats, DB, locales } from '../../db/Database';
+    import { DB, locales, Creators } from '../../db/Database';
     import { isAudience, isFlagged } from '../../models/Moderation';
     import { getUser } from '../project/Contexts';
     import Link from './Link.svelte';
-    import { navigating } from '$app/state';
+    import { navigating } from '$app/stores';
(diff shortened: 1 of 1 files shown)


amyjko (changes requested):
Hi @awang040403. A few notes, similar to the last PR draft you submitted:

* Your code has overwritten the previous implementation of this file, reverting it to Svelte 4 syntax and removing the new chat functionality. Please make your changes on top of the existing functionality, without removing functionality and without reverting to Svelte 4 syntax.
* Your implementation should either 1) avoid using labels like "Owner", or 2) add localization strings so that the labels can appear in all languages, not just English. See the localization guide for how to add strings for new user interfaces.
* The issue requires all collaborators to be shown, but your implementation seems to have that functionality commented out.
* To verify this work, you'll need to start the emulator, create multiple accounts on different private browsing tabs, create a project that has multiple collaborators, and then view the project preview to verify that all creators appear correctly.

If any of the above doesn't make sense, just ask me to explain it, and I can teach you anything you need to know.

Issue #606, opened 2024-11-29
Show creators in project preview


amyjko:
### What are you trying to do that you can't?

When teachers are overseeing a gallery of projects, they need to be able to identify which of their students are working on specific projects. Project previews should show the owner and collaborator usernames, if they are available, in the preview.

### What is your idea?

The basic design would:

- [ ] Update `ProjectPreview` to render owner and collaborator usernames with a `CreatorView`

awang040403:
Hi @amyjko, I talked with you this afternoon and am interested in working on this over the winter break. Could you please assign me to it?

amyjko:
Yes, I can assign you @awang040403. This issue is high priority and needs to be in production by January 17th, so work should begin on it immediately. Comment here if that's not possible.

awang040403:
Hi @amyjko, alright. I will let you know ahead of time I am having trouble finishing it on time. I have a quick question to begin the issue. wordplay-main\src\components\app\ProjectPreview.svelte is the file location that I assume to implement the feature, while would the project preview page be the screenshot or is it the gallery? 
![image](https://github.com/user-attachments/assets/6d3e16e7-989f-4f79-a0a2-12070e799d43)
https://wordplay.dev/galleries Or would it solely be presented in teacher's account?
Thank you for your clarification!

amyjko:
Yes, `ProjectPreview` is where we want to show creators. We want to show them on all pages though, whenever a project has a creator. (It doesn't matter what page the preview appears on).

awang040403:
Hello @amyjko, I have made some progress on editing ProjectPreview, while I am unable to test how my work looks because I do not have access to the firebase on my computer's environment. Therefore, I am unable to see the projects that other people created or save my own work when running wordplay with my own environment. Could you please give me an account or permission with the firebase to test my work? Thank you for your help!
![image](https://github.com/user-attachments/assets/10ec2acb-79db-44be-a659-a62ba4eb199e)

amyjko:
You should be able to run the Firebase emulator following the instructions on the wiki:

https://github.com/wordplaydev/wordplay/wiki/Firebase

Let me know if you run into problems.

awang040403:
Thank you Professor Ko! I will follow the instructions and let you know if I have further issues that I cannot solve.

amyjko:
@awang040403 How are things going on this issue? I saw you closed the draft PR for this. Since this is a critical path issue for Adrienne's class, it'd be great to know the status.

PR #636, opened 2024-12-31
#606: Show the creator and collaborators for projects

awang040403:
# Context

The issue is to show creators and collaborators for projects. For ProjectPreview.svelte, I have imported two new parts in line 5 and added the import of Creators. On line 18, I added the import of a new file ProjectCreatorView.svelte. The new file is for controling the presentating look of the creator and collaborators. On line 82 and 83, I added codes to get the owner and collaborator. I added codes on line 136 to line 164 to show the creator and collaborators. Since for some projects, there may be more than 2 collaborators, which will affect the page UI and make the page distorted, so it will only show two collaborators and add "..." for more than two collaborators.

For CreatorDatabase.ts, I edited lines 69 to 82 to show full username instead of just the first 4 characters of the username. I also created ProjectCreatorView.svelte to edit the presenting look of the owner and collaborators of the projects.

## Related issues

None.

## Verification

I have tested with my own environment and this is the output looks. It works well.
image

## Checklist

This is a full pull request. Is there anything that you would like me to edit?


amyjko:
Hi @awang040403! Thanks for your pull requests. You submitted three, however, all seemingly about different files. There needs to be just one pull request, containing all of your proposed modifications, not three with separate files.

More on pull requests here; it seems like you need to first learn what a pull request is before you submit one:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests

awang040403:
Hello @amyjko, thank you for pointing out my issue! I'm sorry for this inconvenience, and I have assembled my edits into one patch and closed the other pull requests. Please let me know if there are other things that you would like me to edit.

amyjko (changes requested):
Thanks for sharing your first draft! I'm glad you got it to a place where I can see everything and comment. I noticed three problems:

* As part of development, you should always be running `npm run check`, per the wiki. That shows you Typescript and Svelte errors. You have four errors to fix, including some broken imports, an unused import, and a function that you removed from `Creator` that is used elsewhere. You seem to have removed, and created a different function, `getProjectUsername()`, which does a similar thing as the function you removed.
* For some reason, your edits to `CreatorDatabase.ts` have removed all of the comments. Is there a reason you removed them? Please restore them.
* You've created a new `ProjectCreatorView` instead of reusing the existing `CreatorView`. Is there a reason why you didn't reuse `CreatorView` They look nearly identical except for a few minor CSS changes. Always reuse code rather than duplicate it.

awang040403:
Hello @amyjko, thank you for your feedback on my first draft! 

For `CreatorDatabase.ts`, I have restored the getUsername(anonymous: boolean) function that I deleted by mistake when editing my draft with your updated codes. I apologize for deleting all your comments in `CreatorDatabase.ts` because I thought they are the ones I wrote, and I have restored them now. I added the `getProjectUsername(anonymous: boolean)` function to show the full username in projects instead of just the first 4 characters, which will allow better users' experiences.

For the new `ProjectCreatorView.svelte`, I made some differences to the style of the creator and collaborator. There is a happy face for every user, and I have removed it now. It is also created to use the getProjectUsername(anonymous: boolean) function, so the full usernames can be presented. 

This is my first time working on a highly development-related issue with you, so I am worried that if I make too many edits on your code will negatively affect them. So I created a new file, which will affect your codes less. I will also take some time to check `CreatorView` and other related files to see if I could reuse codes instead of creating a file.

Below is the exact same creator and collaborators before adding the new function in `CreatorDatabase.ts` and creating `ProjectCreatorView.svelte`.
Projects

Below is the exact same creator and collaborators after adding the new function in `CreatorDatabase.ts` and creating `ProjectCreatorView.svelte`.
Projects

amyjko:
Thanks for the details. I'm happy to help you learn how to do this work; there's a lot to learn, and it takes time.

A few replies:

* When working on a big software project, duplicating code is a very bad idea. It creates more opportunities for defects, it creates more code to maintain, and it leads to inconsistencies in user experience. That is why there needs to be only one `CreatorView`, and not multiple.
* The goal you're trying to meet, of having a full name shown, instead of just a partial name, makes sense. If you're imagining someone being a teacher and trying to see whose project is whose, this is a useful thing to do. However, doing this breaks a different requirement: preserving the privacy of users, who may expose personally identifiable information in their username (even though we tell them not to). That's why `CreatorView` takes the prop `anonymize`, to allow the person using the view to decide, based on context, whether the username should be anonymized or not. Passing `false` to that prop is the appropriate way to remove anonymize, not copying the file and creating a duplicate `CreatorView` that has a different behavior.
* In copying `CreatorView`, you have removed some its other functionality, like showing the creator's chosen emoji.

This brings us to the hard part: how do you decide when it should be anonymous or the complete name? That depends on context:

* On the projects page, all projects are something the viewer has access to, so it should never be anonymous.
* On the gallery page, it depends on who is logged in: if it is the gallery owner or gallery curator, then they should not be anonymized.

This means that the `ProjectPreview` component needs to have an `anonymize` prop as well, and it should use that to determine whether the `CreatorView` is anonymized. The components that create a `ProjectPreview` (there are many), should pass in an `anonymize` prop based on the context of the page and the current user.

Your current changes make usernames available to everyone, regardless of the context, and regardless of who is logged in.

Does the explanation above help with your next steps? As I see it, you need to:

- [ ] Delete `ProjectCreatorView`
- [ ] Restore the `CreatorDatabase` functionality, deleting your `getProjectUsername()` function, which removed the anonymization functionality.
- [ ] Add an `anonymize` prop to `ProjectPreview`
- [ ] Change all creations of `ProjectPreview` to send `true` or `false` for the new `anonymize` prop, based on who is logged in. As part of this, you'll need to add an `anonymize` prop to `ProjectPreviewSet`, which shows many projects.

Once you do the above, that will restore the original functionality of anonymize usernames for privacy, and I believe finish this issue.

awang040403:
Hello Professor Ko @amyjko, I appreciate your comprehensive feedback! 

-  I have followed your instruction to restore your `src\db\CreatorDatabase.ts` and deleted `src\components\app\ProjectCreatorView.svelte` to avoid redundancy.

-  I tried to edit `ProjectPreviewSet` and add an anonymize prop while there are very few codes related to users, which means I would need to write a lot of codes on my own, and I am worried that my edits as a beginner may lead to large issue to your codes. I noticed that you have written codes in `ProjectPreview` on anonymize, so I would like to use it directly in `ProjectPreview` to determine whether to anonymize or not. 
image

I have added the if statement for creater. If the user is the creater, the full name will be shown.
￼
Pasted Graphic 2

Then, I added default anonymize in the `ProjectPreviewSet`.
￼
Pasted Graphic 3


 Is it okay if I edit it this way? It has functioned well on my end after some tests, and there is lighter edits to your codes.

amyjko:
@awang040403 I understand the fear of making edits that you're not sure about, but that's what software development is. And why we do code review, to give you confidence in your edits. We need to implement this in a way that meets requirements, not in the way that avoids your uncertainty :) You'll be okay: make the edits that implement the needs we have, and I'm here to provide feedback on them to make sure they're correct.

awang040403:
Hello Professor Ko @amyjko, thank you for your encouragement and further clarification on the requirements. I have tried to follow your instructions and wonder if my current work is on the right track to fulfill the requirements. Could you please take a look at it?

I also placed determining whether the user is creator or curator to `ProjectPreviewSet`.
image
image
image

I also commented out determining whether the user is creator or curator to `ProjectPreview`.
image

This is how the website looks like after editing the two files.
image

image

Is this look and functionality okay? If so, I plan to work on other places and add a type of user in Gallery. Thanks again for your patience and continued help!

src/components/app/ProjectPreviewSet.svelte
```diff
@@ -43,12 +45,13 @@
...
 
     let listed = $derived(sortProjects(set).filter((p) => p.isListed()));
+    const user = getUser();//Amy
 </script>
 
 <div class="projects">
     {#each listed as project (project.getID())}
         {@const removeMeta = remove(project)}
-        <ProjectPreview {project} link={project.getLink(true)}
+        <!--amy--><ProjectPreview {project} link={project.getLink(true)} anonymize={isAudience($user, project)}
```
amyjko, line 54: It's closer, but this line is wrong:

* If a `ProjectPreviewSet` is rendered on a gallery page, it should be anonymized only when the user is not a curator or creator on the gallery (because those in a gallery should be able to see others in the gallery). If they aren't, then it should be anonymous.
* If a `ProjectPreviewSet` is on the `/projects` page, it should never be anonymous, because that page shows all of the logged in person's projects. One could argue that they shouldn't show the person logged in though, because all of them would show that.

Because the `ProjectPreviewSet` needs to behave differently depend on where it is shown, it needs an `anonymized` prop as well, and all places they are created should meet the requirements above.

amyjko (changes requested):
It's very close, but the change you made to `ProjectPreviewSet` still doesn't meet the requirement of having the previews be anonymized based on context. Your current version anonymizes all project preview sets in the same way, but they need to be different based on the context in which they are displayed. See the comment on the line where you're passing an `anonymized` value to `ProjectPreview`.

src/components/app/ProjectPreview.svelte
```diff
@@ -138,6 +153,40 @@
                         {project.getName()}{/if}</Link
                 >{#if navigating && `${navigating.to?.url.pathname}${navigating.to?.url.search}` === path}
                     <Spinning />{:else}{@render children?.()}{/if}{/if}
+            <!--add by Amy-->
+            {#if owner}
```
amyjko, line 157: This isn't quite right: collaborators should be able to see who is on the project as well. This check means that only owners will ever see collaborators, but not collaborators, and not gallery curators. This won't meet the requirement of this issue, because teachers who are curators of galleries will not be able to see which students are working on a project.

The way to remedy this is to do a similar thing as with `anonymize`: `ProjectPreview` should have a prop that determines whether to show the collaborators, and that should be determined based on context. For example:

* On the `/projects` page, we should always show collaborators
* On a `/gallery/` page, we should show collaborators if the user is a gallery curator or creator
* For `GalleryPreview`, we should never show collaborators

amyjko (changes requested):
Almost there! There's one critical issue in `ProjectPreview` that prevents the changes from meeting the requirements described in #606. Then, there are many comments you have left that should be removed before the issue is final.

awang040403:
Hello @amyjko Professor Ko, thank you for pointing out the problem and writing detailed instructions. I have edited based on it, and the names are presenting following the instructions. Could you please take a look whether it fulfills the instruction this time?

amyjko (approved):
This looks good! I made a few visual design improvements and removed some unnecessary markup, just to keep things today. I also changed the logic in `ProjectPreview` to only show collaborators if there are more than one, so we don't show the project owner to themselves.

amyjko:
All merged! Congratulations and great job iterating on this one.

awang040403:
Sounds good. Thank you, Professor Ko!

PR #655, opened 2025-01-21
#654: Update SetValue unit tests

### LINKED ISSUE #654: Update one of the unit tests to make it more structured

The current unit tests (vitest) can be structured well. The test can be grouped well using describe and can be written in a more readable way using it blocks.

### What is your idea?

This will ensure we as a team start writing unit tests for both the success and failure cases include edge cases at both ends.

raghavmalawat:
Fixed #654

# Context
Structuring unit tests for `SetValue` in a better way, in order to make it more extendable, and readable. 
This will ensure we can write write all positive, negative, edge test cases and group them of easier understanding, helping us in increasing test confidence.

## Related issues

-   Closes #654 

## Verification
Ran the test locally using `npm run test values/SetValue.test.ts`


raghavmalawat:
@amyjko can you please help me tag the relevant folks here that can help me review the PR.
Also I am unable to add anyone as a reviewer.

amyjko:
I'm the sole reviewer, so you tagged the relevant folks :)

amyjko (changes requested):
This is close; there are a few TypeScript errors to fix first. Always run `npm run check` to monitor for these issues.

raghavmalawat:
@amy I will need some help here in instantiating `Context` instances for the test cases. I have tried reading the docs and checking existing code, but unable to understand how to instatiate it. The test cases pass, it is just that the null should be replaced by an instance of `Context`.

If I can get some help from a developer who has worked on the codebase earlier, I think we should be good here

Screenshot 2025-01-23 at 2 13 59 PM

amyjko:
Thanks @raghavmalawat, I'm the primary developer of Wordplay, so I'm the person to ask.

A `Context` is a set of caches of information about a `Source`, used to support analysis of the source's code. Look at `evaluate.ts` for how to make a test project, then just call `Project.getContext().` For this to work, the value you create should ideally be generated from the `Source` (e.g., make a source with the code `{}`, which produces an empty set value, and then evaluate the project as in `evaluate.ts` to get that final value.

amyjko (changes requested):
This still has a few type errors to resolve. See them in the PR checks, or run `npm run test` to see the issues.

amyjko:
I still see an import issue with import `@models/Project`. (`@models` isn't a valid path alias).

amyjko (changes requested):
Resolve failed `@models/Project` import; that's not a valid path alias.

raghavmalawat:
Hi @amyjko, I have fixed the imports. Also curious, is there a way for us to run the entire pipeline in local setup so that it becomes easier to catch such errors. On running the tests and executing `npx svelte-check --tsconfig ./tsconfig.json .svelte-kit` I couldn't see any local errors, but they surfaced here on Github

amyjko:
Thanks @raghavmalawat. (Did you commit the change? I'm not seeing a new commit on the PR).

The import issues you were seeing weren't visible because you were pointing `svelte-check` to `.svelte-kit`, which only has built files. You're probably seeing them on an old local build, not your fresh source. When you run it with the `--watch` option (as we do with `npm run check`, it uses your current source, not your built source.

Just using VS Code will highlight these as well, without requiring any command line checking.

amyjko (changes requested):
I think the Project import on this is still incorrect.

raghavmalawat:
Screenshot 2025-02-01 at 3 51 12 PM

I am not sure why this is failing, since I am unable to replicate the issue on my local. Additionally I tried importing the `Project` the way it was done in another test file (Basis.test.ts)

amyjko:
Super strange bug; I can't reproduce locally either. Other tests (e.g., `Reaction.test.ts` do the same import and are fine. I'll do some debugging and see what I find.

amyjko:
@raghavmalawat I found the cause:

* Your branch was pretty far behind `main`
* The workflows that run on the PR for approval run on an hypothetical merge with main
* Between the time you created your fork and submitted this PR, the `models` folder was merged with the `db` folder, to tidy up data related files, and so the import was actually incorrect upon merge.
* I updated your branch with the recent changes from main, and the import error showed locally.

I went ahead and fixed the import and merged this PR.

raghavmalawat:
Ah got it, this is a good reminder for me to constantly take a fresh pull from `main`. I remeber reading it on the wiki page but I will be more mindful about it. I knew it might be something silly that I adoing, but had spent good amount of time figuring it out, before asking for help. Thanks for unblocking me here @amyjko

PR #674, opened 2025-02-08
Added large deletion warning popup above experimental message

nathanbek:
# Context

## Related issues

-   Related Issue # 499
-   Closes #

## Verification

 Verified on Chrome and Safari that when a large deletion occurs, the notification pops up with the desired black background, white text, and smooth pop-up animation.
• Confirmed that triggering undo via the undo button or pressing Ctrl/Cmd+Z immediately clears the notification.
• Checked that subsequent edits clear any existing notification as expected.
Accessibility:
• The notification message reads clearly with white text on a black background.
• The pop-up animation is subtle and does not impede interaction.

## Checklist

[x] Inline the large deletion notification store in Editor.svelte
[x] Changed the notification message to display the actual arrow icon (↺).
[x] Updated ProjectView.svelte to import the inline store & apply custom styling (black background, white text, 1px solid black border, pop-up animation).
[x] Added logic to clear the notification on an undo event.

Screenshot 2025-02-07 at 8 40 45 PM

### DIFF
--- a/src/components/editor/Editor.svelte
+++ b/src/components/editor/Editor.svelte
@@ -20,6 +20,7 @@
     import Sym from '@nodes/Sym';
     import Token from '@nodes/Token';
     import TypePlaceholder from '@nodes/TypePlaceholder';
+    import { UNDO_SYMBOL } from '@parser/Symbols';
     import type Evaluator from '@runtime/Evaluator';
     import ExceptionValue from '@values/ExceptionValue';
     import { onMount, tick, untrack } from 'svelte';
@@ -82,6 +83,9 @@
     } from './util/Highlights';
     import Menu, { RevisionSet } from './util/Menu';
     import { OutlinePadding } from './util/outline';
+
+    // Local deletion notification state (instead of a global store)
+    let largeDeletionNotification: string | null = $state(null);
 
     interface Props {
         /** The evaluator evaluating the source being edited. */
@@ -394,7 +398,7 @@
 
     function handlePointerDown(event: PointerEvent) {
         // Clear any existing large deletion notification when user clicks to clear selection
-        largeDeletionNotification.set(null);
+        largeDeletionNotification = null;
         event.preventDefault();
         event.stopPropagation();
 
@@ -953,7 +957,7 @@
         if (edit === undefined) return;
 
         // Clear any existing large deletion notification since a new edit has started
-        largeDeletionNotification.set(null);
+        largeDeletionNotification = null;
         const previousSource = source;
 
         const navigation = edit instanceof Caret;
@@ -1025,8 +1029,8 @@
                 newSource.getCode().getLength() >=
                 40
         ) {
-            largeDeletionNotification.set(
-                'Are you sure you want to delete this selection? You can use the undo button (↺) if you change your mind.',
+            largeDeletionNotification = $locales.get(
+                (l) => l.ui.source.cursor.largeDelete,
             );
         }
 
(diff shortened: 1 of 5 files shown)


amyjko (changes requested):
This is a great start! A few minor revisions:

* You used a global store to store the deletion notification. It should instead be a local `$state`, one per `Editor` component, so if there are multiple editors, there is one notification state per editor. The way to pass this state to `ProjectView` for display is to use the Svelte `bind:` feature, which allows to binding to an array, binding to one state per editor, and then displaying state based on that editor.
* The notification text needs to be localized. See the various other patterns that use the `$locales` store to get the locale string. You'll need to update `UITexts.ts`, adding a new locale string to the `cursor` object (e.g., `largeDelete`), then run `npm run schemas` to generate the new schema, update `en-US.json` with the new string text. I can run `npm run locales translate` to generate machine translations for all of the other supported languages.
* The undo capture is hacky, duplicating keyboard shortcut checks. The undo command is stored in `Commands`, named `Undo`. The way to build this within the current design patterns is to update that command to do the resetting work, or check after the call to `handleKeyCommand` to see if the Undo command was the one processed.
* The notification view is nice, including the animation. But it's intended to follow the same visual design as the `editor-warning`, so you should merge these to CSS configurations into one class, and reuse the class, rather than duplicating the CSS.
* We have a global animation control for controlling animation speed, for accessibility purposes. Instead of hard-coding an animation speed in the popup, refer to the `var(--animation-factor)` CSS variable. Search for other uses to see how this is used to speed up and slow down animations.
* The animation speed should be consistent to other animation speeds in the application. Most are anchored at 250ms, not 0.6s.

Resolve these issues and it should be ready to go. Nice work!

nathanbek:
Hey @amyjko 
Here are some things I worked on as requested:

• The global deletion notification store was removed. Now each Editor instance gets its own reactive deletion notification state via $state, supplied through $props() so that ProjectView can bind an array of these states.

• The notification text is localized. I updated UITexts.ts to add a new property (e.g., “largeDelete”) to the cursor object and updated en‑US.json with the appropriate string so that the deletion message is pulled from $locales in the code.

• The undo capture hack was addressed by checking after handleKeyCommand if the processed command is Undo (using the UNDO_SYMBOL) and clearing the deletion notification accordingly.

• The notification view now uses the merged CSS class (editor‑warning) so that its design is consistent across the app. The animation timing now uses var(--animation-factor) with a 250ms base duration to match other animations in the application.

src/components/editor/Editor.svelte
```diff
@@ -83,6 +84,9 @@
     import Menu, { RevisionSet } from './util/Menu';
     import { OutlinePadding } from './util/outline';
 
+    // Local deletion notification state (instead of a global store)
```
amyjko, line 87: State should be declared below the props interface, and preferably nearby other related state, as the component has become quite large. Organization really matters here, to help others make sense of it.

src/components/editor/Editor.svelte
```diff
@@ -1197,26 +1180,23 @@
             getTokenViews,
         });
 
-        // Don't insert symbols if composing.
-        insertedSymbol = command === InsertSymbol;
+        // If the command processed is the Undo command, clear the deletion notification
```
amyjko, line 1183: If the commands symbol changes for some reason, this will break. Better to check to see if the Undo command itself was selected, rather than checking its symbol.

src/locale/en-US.json
```diff
@@ -4314,7 +4314,8 @@
                 "redo": "redo undone edit",
                 "search": "search for special characters to insert",
                 "tidy": "tidy spacing",
-                "elide": "toggle elision"
+                "elide": "toggle elision",
+                "largeDelete": "Are you sure you want to delete this selection? You can use the undo button (↺) if you change your mind."
```
amyjko, line 4318: The phrasing of this is a bit off, because the interaction is that the delete happens, and then they have the option of undoing. It should be in past tense, not present tense, since the deletion already happened. Maybe something like:

"That was a big deletion. You can undo (↺) if it was an accident."

The other issue here is that the undo symbol could change in the future. One alternative (and improvement) to putting it inline here is to put an undo `CommandButton` after this text, instead of making them hunt for the button in the toolbar.

amyjko (changes requested):
Very close! I left a few comments for minor improvements, and one slightly bigger one, but still minor.

src/components/editor/Editor.svelte
nathanbek: I've updated the code to check for the Undo command object itself (if (command && command === Undo)) rather than checking its symbol, making it more resilient to future changes.

src/locale/en-US.json
nathanbek: I've updated the notification text to use past tense as suggested: "That was a big deletion. You can undo if it was an accident." Also, I've implemented the CommandButton directly in the notification itself rather than requiring users to hunt for it in the toolbar.

amyjko (approved):
Approved! Congrats on this one, it turned out well.

amyjko (changes requested):
Actually, there are still a few merge conflicts here to resolve. Resolve those and I'll merge and deploy.

Issue #629, opened 2024-12-24
Type checking in Firebase cloud functions


amyjko:
### What are you trying to do that you can't?

We define many callable Firebase Cloud functions in `functions/src/index.ts`, but the types are often not declared and types are not shared between the client and the cloud functions. Instead, we're currently defining them in `/src/db/functions.ts` and `/functions/src/types.ts` and copying them manually. Yuck.

### What is your idea?

There are ways of defining shared libraries and allowing them to be used on both client and server. Explore these possibilities, define the correct types for all functions, fix any defects we find, and document how it works.

annieli555:
Hi @amyjko! I'm interested in working on this issue. May I be assigned?

amyjko:
Assigned! This one requires some research on best practices and then an implementation. I found a few online, but they were old, so I'm not sure if there are better solutions.

annieli555:
Hi Amy!

Apologies for the delay here. I investigated sharing types between cloud functions and the client, and found a few common solutions mentioned. I've opened [a PR](https://github.com/wordplaydev/wordplay/pull/694) implementing option 1 because it felt the cleanest, but am listing them all here for completeness.

1. **Set up a local npm package**. This approach was suggested by this [answer in Reddit](https://www.reddit.com/r/Firebase/comments/xozd0s/comment/iq4fi7t/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button). It sounded like others in the thread had trouble getting this to work in their environments, but I was able to get the types to resolve locally.

2. **Copy the file during build time** (settled on by OP in the same [thread above](https://www.reddit.com/r/Firebase/comments/xozd0s/does_anyone_have_an_example_of_using_shared_types/)). This feels hacky to me, but seems like an option that we could do as a last resort.

3. **Use a symlink**. Unfortunately, I wasn't able to get this to work (I might have misconfigured something - I tried linking wordplay src from functions and wasn't able to get the reference to src/db/functions to resolve). Also, according to this [Stack Overflow post](https://stackoverflow.com/questions/66239777/deploying-firebase-functions-with-local-depencies-using-firebase-cli), they tried something similar under "Approach 1" but got errors trying to deploy functions.

[...]

PR #694, opened 2025-02-22
Share types between client and cloud functions through shared package

annieli555:
# Context

Sets up a local package under the `functions` directory that contains types to be shared between cloud functions and the client. This approach follows a recommendation in [this Reddit thread](https://www.reddit.com/r/Firebase/comments/xozd0s/comment/iq4fi7t/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).

This approach should "just work", and any edits to the types in `functions` should be reflected in `src` (I've tested this locally by editing `functions/src/shared/index.ts` such that it triggers type errors in `src/db`).

## Related issues

Closes https://github.com/wordplaydev/wordplay/issues/629

## Verification

Verified that build succeeds locally and type errors are reflected between packages without having to build.


amyjko (changes requested):
This looks like a great start, and it seems quite close. The main issue I'm seeing is that it breaks the dev workflow with vite, since the only time shared types are updated is on `build`. When the vite dev server is running, if you edit a type, it won't update shared types, requiring a manual build. There may be some ways to use vite hooks to include this in any updates to dependent files. Look into that and see what's possible?

annieli555:
Hey Amy! Sorry for the delay here. I finally got the chance into this more and started writing a Vite plugin that watches for changes in the shared types directory and triggers a full reload [0].

However, I'm a little confused what the goal is by rebuilding the package when the dev server is running - could you help me understand more what you're looking for? When I make changes locally to the shared types under `functions/`, I see type changes / errors reflected in `src/` [1]. However, the vite server won't crash or reflect any errors if I update the types, (and it looks like this was true before this change as well - if I update `src/db/functions.ts`, the dev + build commands still succeed).

[0] Something like this:
```javascript
/**
 * Plugin to react to changes in the shared types package containing common types
 * between cloud functions and client code.
 */
function watchSharedTypes () {
    return {
        name: 'watch-shared-types',
        configureServer (server) {
            const sharedTypesDir = path.resolve('./functions/src/shared');

            fs.watch(sharedTypesDir, { recursive: true }, () => {
                server.ws.send({ type: 'full-reload' });
            });
        }
    };
}
```

[1] If I remove 
Screenshot 2025-03-07 at 8 00 27 PM


In the `src/db/creators` file:
Screenshot 2025-03-07 at 7 59 42 PM

vite.config.js
```diff
@@ -21,6 +21,11 @@ function LocaleHotReload() {
 const config = {
     plugins: [sveltekit(), LocaleHotReload()],
     build: { chunkSizeWarningLimit: 1600 },
+    server: {
```
annieli555, line 24: While testing, I encountered a `The request url "/Users/annieli/Development/wordplay/functions/src/shared/index.ts" is outside of Vite serving allow list.` error when I tried to navigate to the login page. This new entry fixes that.

src/db/functions.ts
```diff
@@ -1,32 +0,0 @@
-// WARNING: We haven't set up a shared library for these types, so these must be synchronized between
```
amyjko, line 1: I believe this comment is obsolete after this PR. Delete if so.

amyjko (changes requested):
One tiny little comment about a comment, but after that's resolved, this is ready to ship!

src/db/functions.ts
annieli555: This file is being deleted! But good catch, before my last commit I hadn't deleted the comment in the shared types file. Now it is gone :)

https://github.com/wordplaydev/wordplay/pull/694/files#diff-a76a2413f6b49c49a2f7ada05258ff52b91ca337f5b8c95b2c6c2b4d878096bdL1-L2

amyjko (approved):

amyjko:
Okay, all merged. Congratulations!

PR #726, opened 2025-03-24
Fix/sequence description

nathanbek:
# Context

This PR adds the `getDescription` method to the `Sequence` class, fixing an issue where the absence of this method caused errors when other components tried to access it. This issue was causing placeholder symbols (⚡⚡ and $?) to appear throughout the UI instead of proper localized text.

## Related issues

- Related Issue #725

## Verification

I've verified these changes by:

1. Adding the `getDescription` method to the `Sequence` class that returns the description of the first pose or an empty string if no poses exist
2. Building the project successfully with `npm run build`
3. Testing in Chrome browser that:
   - UI displays correctly without placeholder symbols ($? and ⚡⚡)
   - Animations render properly
   - No infinite locale refresh loops occur

This fix resolves issues encountered in PR #725 that caused continuous locale file regeneration and browser refresh loops.

## Checklist

- [x] Added missing `getDescription` method to `Sequence` class
- [x] Verified the build completes successfully
- [x] Tested that UI displays properly without placeholder symbols
- [x] Ensured no changes to vite.config.js or locale files were included in this PR


amyjko (changes requested):
Looks like there's a conflict to resolve in this one before it's ready.

PR #735, opened 2025-04-10
Created guide for styling text

elainel7:
# Context

## Description

This PR adds a new guide for styling text. It describes how to italicize, bold, and underline text.

## Related issues
None.

## Verification

I verified by reviewing the curate guide:
1. Added a styling-text.txt file in `static/locales/en-US/how`.
2. Added an identifier to it in `HowTo.ts`. 

## Checklist

- [x] Make sure there isn't an 'how to' for the design pattern already
- [x] Create a file in `static/locales/en-US/how` (or any other locale if you want to create a locale-specific example). You'll need to choose a unique identifier for it.
- [x] Add that identifier to `HowTo.ts`, which has an object that should contain the ID. Specify a valid category for the design pattern.
- [x] Write the how to document in your text file. The first line should be a title, the last line one more or related how to IDs (at least one is required), and everything else in between should be Wordplay markdown syntax. Follow the other examples for the tone and content.


amyjko (changes requested):
This is a great start! I think it's close, but needs a few revisions:

* It's not `Scene` that supports formatting, it's `Markup`, and currently `Phase` can take a `Markup` and render it. Referring to `Phrase` or `Markup` would be more accurate here, as `Scene` has nothing to do with markup.
* There's a lot of documentation in `Markup` and `Phrase` about these options already, so this is a bit redundantly. Could you think of a way to make this how to go a bit beyond what the basic documentation for `Markup` does? One idea might be to provide some example context where rich text might be really helpful, like creating a paragraph of text with rich text, maybe even with links.

amyjko (approved):

PR #741, opened 2025-04-22
Add description to browser selection in defect issue template

### LINKED ISSUE #714: Unable to Select All Browsers
### What happened?

Image

Image

Right now there’s no option to select all browsers at once. If the issue isn’t specific to a certain browser, it’s hard to highlight it and currently that can’t be done. This makes it difficult to address problems that might be affecting multiple browsers. Having this option would make it easier to spot and fix issues that aren't tied to just one browser.

### Anything else?

_No response_

Chrome

macOS

annieli555:
# Context

The dropdown field in the browser selection prompt allows the reporter to select multiple options, but the GitHub UI doesn't make that obvious. Add a description that clarifies that the reporter needs to select options one by one if they wish to specify multiple browsers.

## Related issues

-   Closes #714

### DIFF
--- a/.github/ISSUE_TEMPLATE/defect.yml
+++ b/.github/ISSUE_TEMPLATE/defect.yml
@@ -32,6 +32,8 @@ body:
       id: browsers
       attributes:
           label: What browsers are you seeing the problem on?
+          description: |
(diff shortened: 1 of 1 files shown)


amyjko (approved):
Thanks for this improvement.

PR #743, opened 2025-04-23
665: Project previews should show custom emoji

lantinglu:
# Context

This pull request addresses an issue in the ProjectPreview component where custom emoji characters were not being displayed properly. The bug was caused by incomplete parsing of character names from StructureValue or MarkupValue instances, which led to fallback rendering instead of the correct emoji SVG.

## Related issues

-   Related Issue # 665 Project previews should show custom emoji

## Verification

npm run dev-Manually tested in Chrome
Checked that regular character previews continue to render properly.

## Checklist

Fixed character preview rendering issue in `ProjectPreview.svelte`
Updated `findCharacterName` to support `StructureValue`
Verified custom emoji characters display correctly
Manually tested in Chrome

### DIFF
--- a/src/components/app/ProjectPreview.svelte
+++ b/src/components/app/ProjectPreview.svelte
@@ -16,6 +16,11 @@
     import CreatorView from './CreatorView.svelte';
     import Link from './Link.svelte';
     import Spinning from './Spinning.svelte';
+    // new added
+    import type { Character } from '../../db/characters/Character';
+    import { characterToSVG } from '../../db/characters/Character';
+    import ConceptLink, { CharacterName } from '../../nodes/ConceptLink';
+    import MarkupValue from '../../values/MarkupValue';
 
     interface Props {
         project: Project;
@@ -29,6 +34,42 @@
         children?: import('svelte').Snippet;
         anonymize?: boolean;
         showCollaborators?: boolean;
+    }
+
+    function findCharacterName(value: unknown): string | null {
+        // If it's a MarkupValue, check for character links
+        if (value instanceof MarkupValue) {
+            const nodes = value.markup.nodes();
+            for (const node of nodes) {
+                if (node instanceof ConceptLink) {
+                    const parsed = ConceptLink.parse(node.getName());
+                    if (parsed instanceof CharacterName) {
+                        return `${parsed.username}/${parsed.name}`;
+                    }
+                }
+            }
+        }
+
+        // If it's a StructureValue, check all its fields recursively
+        if (
+            value &&
+            value.constructor &&
+            value.constructor.name === 'StructureValue' &&
+            (value as any).context &&
+            (value as any).context.getBindingsByNames
+        ) {
+            const structureValue = value as any;
+            // Get all bindings from the context
+            const bindings = structureValue.context.getBindingsByNames();
+
+            // Loop through each binding and check if it contains a character name
+            for (const [_, fieldValue] of bindings) {
+                const result = findCharacterName(fieldValue);
+                if (result) return result;
+            }
+        }
+
+        return null;
     }
 
     let {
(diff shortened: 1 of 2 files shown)


lantinglu:
@amyjko Hi Amy, could you please review this pull request?

src/components/app/ProjectPreview.svelte
```diff
@@ -56,6 +99,8 @@
         representativeBackground,
         representativeFace,
         representativeText,
+        isCustomCharacter,
```
amyjko, line 102: Why is this field necessary? It seems like it's redundant with characterName, which is null if none was found. I recommend remove it, and just using the name.

src/components/app/ProjectPreview.svelte
```diff
@@ -31,6 +36,42 @@
         showCollaborators?: boolean;
     }
 
+    function findCharacterName(value: unknown): string | null {
```
amyjko, line 39: The argument `value` shouldn't have an unknown type. It is of type `Value`.

src/components/app/ProjectPreview.svelte
```diff
@@ -85,9 +145,43 @@
...
+    // Effect to fetch the character from the database when characterName changes
+    $effect(() => {
+        // Clear character if characterName is not valid
+        if (!isCustomCharacter || !characterName) {
+            character = null;
+            return;
+        }
+
+        // Make sure DB.Characters exists and has getByName method
+        if (!DB.Characters || typeof DB.Characters.getByName !== 'function') {
```
amyjko, line 165: Why is this check for `getByName` here? It is always defined and always a function. The same is true of `DB.Characters`. Why is either check necessary?

src/components/app/ProjectPreview.svelte
```diff
@@ -85,9 +145,43 @@
...
+        DB.Characters.getByName(characterName)
+            .then((result) => {
+                if (result) {
+                    character = result;
+                } else {
+                    character = null;
+                }
+            })
+            .catch((error) => {
+                console.error('bug', error);
```
amyjko, line 180: This is not a suitable error message. If there was an error loading the character, there should be a more descriptive console message explaining the context of what was happening (e.g., the function already returns null or undefined for different not found conditions, so this is a more fatal error, like the connection to Firebase being broken, or there being a permissions error).

amyjko (changes requested):
This is very close, nice work on getting this to functional. Just a few minor things polish; let me know if you have any questions?

src/components/app/ProjectPreview.svelte
```diff
@@ -31,6 +37,42 @@
...
+                }
+            }
+        }
+
+        // If it's a StructureValue, check all its fields recursively
+        if (
+            value &&
+            value.constructor &&
+            value.constructor.name === 'StructureValue' &&
+            (value as any).context &&
```
amyjko, line 59: Never use `any` in TypeScript. This expression should be checking if the value is an instance of `StructureValue` (e.g., "value instanceof StructureValue"). That is the type safe way to verify a value's type.

src/components/app/ProjectPreview.svelte
```diff
@@ -31,6 +37,42 @@
...
+            (value as any).context.getBindingsByNames
+        ) {
+            const structureValue = value as any;
```
amyjko, line 62: No `any`. The `instanceof` I recommend above will type guard for the type.

amyjko (changes requested):
A few more typescript problems, then this should be ready.

src/components/app/ProjectPreview.svelte
lantinglu: I have fixed this. Could you please review the changes?

amyjko (approved):
Approved! Congratulations. This will be live in production soon.

lantinglu:
Thank you Amy!

PR #765, opened 2025-05-15
Layering images

elainel7:
# Context

## Description

This PR adds a new guide for layering shapes/images. It describes how to create new images by using multiple shapes.

## Related issues
None.

## Verification

I verified by reviewing the curate guide:
1. Added a `layering-images.txt` file in `static/locales/en-US/how`.
2. Added an identifier to it in `HowTo.ts`. 

## Checklist

- [x] Make sure there isn't an 'how to' for the design pattern already
- [x] Create a file in `static/locales/en-US/how` (or any other locale if you want to create a locale-specific example). You'll need to choose a unique identifier for it.
- [x] Add that identifier to `HowTo.ts`, which has an object that should contain the ID. Specify a valid category for the design pattern.
- [x] Write the how to document in your text file. The first line should be a title, the last line one more or related how to IDs (at least one is required), and everything else in between should be Wordplay markdown syntax. Follow the other examples for the tone and content.


amyjko (commented):
This looks like a solid draft. Are you requesting feedback on it yet, or still working on it? It looks good enough to me to publish. Mark it ready for review if it's ready.

amyjko (approved):
This looks ready to go to me! Nice work, I think it's a great example of something between API and project example.

Issue #641, opened 2025-01-13
Design pattern templates guide


amyjko:
### What are you trying to do that you can't?

There are many design patterns for writing Wordplay programs that are important for achieving different behaviors (e.g., showing a sequence of animated scenes, a sequence of interactive scenes, moving a character around), and these are demonstrated throughout the example galleries and the tutorial.

However, they aren't *explicitly* explained anywhere. This makes it hard for creators to map questions like "how do I do X?" onto a template for code that does X, because it requires scouring all of the examples and inferring the template from the concrete example.

### What is your idea?

The high level design concept is to add a section to the Guide that does the following:

- [ ] Provides a set of explicated example projects, each illustrating a particular design pattern
- [ ] The set of design patterns should be extracted from the existing examples in the galleries and tutorials 
- [ ] The templates have titles that explicitly answer "how do I" questions (e.g., "move a character")
- [ ] Add these templates to the new project dialog, possibly enhancing the dialog with some basic search features

We need to design and build this, ideally before mid-February, in time for Adrienne's class.

@AfifahK, this was the issue I was talking about. This one is high priority; let's talk about it and see if we can make some progress on design!

AfifahK:
thoughts:
- search, system of organization in templates
- common design patterns (pull ideas from scratch, code.org)
- prioritization of which to create first! (let's create brainstorm list at meetup 1/22)

MadhumithaSridhar:
Notes/ideas:

1. Possibly organize into categories like interactive scenes, animations, character movements. This can be broken into subcategories (e.g. looped animations, event-driven interactivity)
2. We can maybe use question-based titles for templates like: "How do I move a character?", "How do I animate a background?", etc.
3. pulling inspo from Scratch/code.org like @AfifahK sounds like a great idea -- things we can possibly include could be event listeners (e.g., keypress for movement), broadcast and receive (e.g., transitioning between scenes), and loops and conditions (e.g., animations that pause after a condition is met). 
4. potential ideas for templates: 

- character movement:  E.g., move left/right, jump, or follow a path.
- Scene Transitions: E.g., fade in/out, slide to next scene.
- Interactive Choices: E.g., buttons for player decisions.
- Timed Animations: E.g., sprite animation loops.
- Dialogue Systems: E.g., NPC interactions with text.

elainel7:
- Link from guide to existing templates to help find examples of full examples of using a design pattern
- Make mini templates with snippets of how to use a design pattern 
[...]

PR #766, opened 2025-05-15
Created guide on repeating character movements

navyacodes:
# Context

This pull request adds a new guide for repeating character movements. It describes how to make a character move a certain number of times. 

## Related issues

-   Related Issue #641 

## Verification

I verified my changes by reviewing and following the curate guide, and running my copy on my browser (Chrome) to ensure it works. 

## Checklist

- [x]  Make sure there isn't an 'how to' for the design pattern already
- [x] Create a file in static/locales/en-US/how (or any other locale if you want to create a locale-specific example). You'll need to choose a unique identifier for it.
- [x]  Add that identifier to HowTo.ts, which has an object that should contain the ID. Specify a valid category for the design pattern.
- [x]  Write the how to document in your text file. The first line should be a title, the last line one more or related how to IDs (at least one is required), and everything else in between should be Wordplay markdown syntax. Follow the other examples for the tone and content.


amyjko (approved):

amyjko (commented):
This looks like a good addition!

PR #793, opened 2025-06-09
582-fr-FR

EddieNew:
# Context

The French localization (France dialect) was initially machine translated. I sat down, researched, and determined the right level of formality and specific phrasing that would make the machine translation more friendly toward our intended user base of diverse children (i.e. deciding on a more casual tone, using gender-neutral terms where possible, adding some basic and common slang, etc.). I then implemented these decisions for the vocabulary in the main file and the beginning of the tutorial (up to about line 200, so about 10% done) I also proofread grammar and made minor edits in the tutorial as well. I had another friend review my updates, though more eyes to proofread are always appreciated!

# Next Steps

There is still a big chunk of the tutorial to go - I will leave some general tips to maintain a consistent localization (backed by my research!). And if willing, proofread my earlier edits.

# For Later Users

For the formality, I had the functions refer to each other and the director with the informal "tu" to maintain a more friendly and casual tone. If you wanted to continue localizing the site that would be an easy place to start (replacing all the "vous" and respective verbs). Also, to avoid gendered language when describing the user, I've been using the slang "le boss" in place of a more formal term (the word itself is still technically masculine, but it is used for all genders without editing, unlike director or administrator). If applicable, the gender neutral pronoun "iel" is recommended (its conjugations are the same as il or elle). Finally, if you cannot avoid using a gender neutral term, I would suggest using the dot formation to show both gendered versions: i.e. un·e éditeur·rice.


amyjko (changes requested):
Thanks for the improvements to the French locale! One note on these before I merge: it looks like several of the `fr-FR` edits keep the machine translated annotation (`$~`). Did you mean to leave those there? They should be removed if they are written by a human, so we don't override them with later machine translations.

EddieNew:
Shoot yeah, those lines must have slipped my mind. I'll fix that now!

EddieNew:
Should be fixed - I'm not sure how this works so let me know if I need to submit another pull request.

Issue #811, opened 2025-06-21
Evaluate autocomplete is flaky


amyjko:
### What happened?

It simply doesn't work as intended. Consider this situation, where `*` represents the caret position:

`Phrase(*)`

Pressing the autocomplete menu or keyboard shortcut correctly recommends an empty string, but:

* Doesn't recommend a formatted literal
* Recommends optional inputs, even though inserting them would be invalid, since the required text wouldn't exist

Then, in this situation:

`Phrase('' name: '' *)`

* No other optional inputs are recommended

And in this situation

`Phrase('' * name: '')`

* Inputs *are* recommended, but are inserted at the end instead of where the caret is.

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

macOS

PR #858, opened 2025-11-15
Fix menu interaction & cleanup logic

BoyanLu1008:
# Context

Issue #811 reports that autocomplete is "flaky" when filling function parameters in Evaluate nodes. Two main problems - optional parameters showing up before required ones in suggestions, and caret jumping to random positions after inserting parameters.

## Related issues

-   Related Issue #
-   Closes 811#

## Verification
Added filtering logic to prioritize required parameters in autocomplete. When there are unfilled required params, optional ones get filtered out. Did this in three places - Evaluate.getPossibleEvaluations(), Autocomplete.getRelativeFieldEdits(), and Autocomplete.getPossibleNodes().
Also fixed the caret position issue in Assign.ts - now it stays at the insertion point instead of jumping to placeholders unless it's explicitly a completion operation.
Tested on Chrome with basic function calls that have required and optional parameters. Required params show up first, optional ones stay hidden until all required are filled. Caret stays where you're typing.
Haven't fully tested variable-length params or nested Evaluate nodes yet. 

## Checklist
- [x] Implemented parameter filtering logic in Evaluate.ts and Autocomplete.ts
- [x] Fixed caret position behavior in Assign.ts
- [x] Tested basic function calls with mixed required/optional parameters
- [ ] Need feedback on the three-layer filtering approach
- [ ] Test edge cases (variable-length params, nested evaluates)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "wordplay",
-    "version": "0.6.25",
+    "version": "0.6.26",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "wordplay",
-            "version": "0.6.25",
+            "version": "0.6.26",
             "bundleDependencies": [
                 "shared-types"
             ],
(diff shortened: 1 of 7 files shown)


package-lock.json
amyjko: Please resolve this conflict. There shouldn't be a change to `package-lock.json`.

PR #875, opened 2025-12-12
Add feature: Choose language upon first visit

### LINKED ISSUE #775: Choose language upon first visit

Non-English fluent visitors can't easily find how to change the language because we default to English.

### What is your idea?

Show a prompt for first time visitors. We need to design this UX: would it use the existing language chooser, or would it be a custom first time visitor screen?

hana-amos:
# Context

The issue addresses the difficulty non-English fluent visitors face in locating the language settings, since the site defaults to English. This PR adds a clear language prompt for first-time visitors, showcasing Wordplay's multilingual options and making it easy for users to select their preferred language right away.

Note: 

- I implemented the hovering state for the rotating locale button outside of the Button component, so that hovering pauses the rotation. This behavior isn’t handled within the Button template itself. I’m wondering if this approach is acceptable or if there’s a better way to handle it.
- Added $! to the l.ui.page.landing.value because it was modified (added "Get started in ")
- "or choose" and "other languages" were marked as unwritten strings ($?)

## Related issues

-   Related Issue #775
-   Closes #775 

## Verification
I have tested my changes for functionality on Google Chrome and Microsoft Edge.

1. Open Wordplay's landing page.
3. Confirm that the language prompt with 2 buttons appears in the speech bubble.
4. Hover over the rotating locale button to verify if it pauses, the correct tooltip appears, and that clicking it changes the language to that locale correctly across Wordplay.
5. Click the "other languages" button to verify if it opens the locale menu (same functionality as locale menu button in footer), and that you can select different languages.

The 2 buttons should have aria-labels (built into Button.svelte).

## Checklist
- [ ] Iterate based on feedback
- [ ] Confirm that the TypeError which comes up from file Webpage.ts only applies to my machine

### DIFF
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -99,8 +99,8 @@
                         class="locale-button-wrapper"
                         onmouseenter={() => (isHovering = true)}
                         onmouseleave={() => (isHovering = false)}
-                        onfocus={() => (isHovering = true)}
-                        onblur={() => (isHovering = false)}
+                        onfocusin={() => (isHovering = true)}
+                        onfocusout={() => (isHovering = false)}
                     >
                         {#key rotatingLabel}
                             <Button


src/routes/+page.svelte
```diff
@@ -195,6 +271,49 @@
...
+
+    .welcome :global(.message) {
+        line-height: 1.7;
+    }
+
+    .locale-button-wrapper {
+        display: inline-block;
+    }
+
+    :global(button.locale-picker.background.padding) {
```
amyjko: This feels a bit hacky, and the tighter spacing for the buttons looks reasonable. I recommend removing this and just changing the padding for all buttons with backgrounds in Button.svelte.

amyjko (changes requested):
This is very close, nice work! Just two minor issues I noted, one related to keyboard accessibility, and one related to hacky CSS workarounds. I also added machine translations for other locales. Clean up those two issues and this is ready to go.

Issue #819, opened 2025-08-26
Finer-grained collaborator roles and corresponding functionality for galleries


melissachen2000:
### What are you trying to do that you can't?

From co-design session with 8th grade students about sharing: 

Students and teachers are interested in having more fine-grained control over who can interact with their project and how. My understanding is that currently, if a student's project is in the class gallery, everyone in the class is able to see it. Additionally, the teacher is able to use the chat feature to have a conversation with the student about their project. However, while students can easily see each others' projects, they cannot easily have a conversation about it in Wordplay if the other student is not added as a collaborator. Students may not want to add others as collaborators, since this also includes editing powers. This means that the platform does not support peer feedback (e.g., a gallery walk, swap and give feedback) yet.

### What is your idea?

The 8th grade students' ideas include:

1. A way to make a project private within the gallery, either sharing only with the teacher / gallery curator or with the teacher and a few select peers (teacher-only viewing in-progress in #738 )
2. More fine-grained roles for collaborators. Introduce a new "reviewer" or "commenter" type (like in Google Docs or Canva) where the collaborator is able to offer suggestions or feedback but cannot directly edit the project. Introduce a new "viewer" type where the collaborator can only view the project, but cannot edit or suggest changes.
4. Chat access for non-collaborators. Allow reviewers and viewers to also be able to chat with the project creator in Wordplay.
5. If 3 is implemented, it will be useful for students to be able to specify what feedback is wanted, turn comments on or off, and moderate comments. Moderation may mean that the teacher approves all comments, or that students have a way to flag or report comments to the teacher so that the teacher may have a conversation about how to give constructive feedback.

amyjko:
Thanks @melissachen2000! This definitely intersects with #738, but it's a broader scope, so let's keep them both and linked.

* From a design perspective, I think the challenges here are in communicating the behavior of these more fine-grained controls. There's also lots of design work to do on the commenting features, especially around moderation.
* From an implementation perspective, I think the challenge is working around Firestore and it's fairly strict limits on its rule queries. These behaviors are *probably* expressible, but I regularly run into try things that are impossible to express without some error-prone data duplication in documents. We should definitely finalize the data model and verify it's expressibility before implementing.

amyjko:
[...]

PR #926, opened 2026-01-26
Adding moderation functionality to chats in galleries

melissachen2000:
# Context

Adds moderation features to chats that are attached to a project or how-to that is in a gallery (private and public galleries). 

See [issue 819](https://github.com/wordplaydev/wordplay/issues/819#issuecomment-3665990199) for lightweight moderation proposal. 
- Namely, note that all moderation action is taken offline. The user who sent the message or reported it is not notified in this current implementation; only the gallery curator, who assumes the responsibility of moderator, is notified. 
- Furthermore, moderation action is only available on chats that are part of a gallery; if a project is not in a gallery, then its chat does not have moderation features. 

I also added a line to the "Rights" page trying to explain chat moderation as separate, but I worry it clashes with the "we only moderate public projects." Are there any edits I can make to help clarify this? 

## Related issues

-   Related Issue #819 
-   Closes #

## Verification

1. Create three test accounts, users A, B, C.
2. Create a gallery using user A's account. Add users B and C as creators. Create a project in the gallery using user B. Add user C as a collaborator to the project.
3. Message from user C. User B should see the report button appear below the message. Report the message. 
4. User A should receive a notification saying that a message requires moderation. Click the link in the notification to go to the moderation dashboard. Clicking the "view" button should take user A to the project. The message that was reported has a button below it that takes the user back to the moderation page. User A should be able to see the reported message. Users B and C should both see placeholder text explaining that the message was removed pending moderation action.
5. User A removes the message via the dashboard. All users should see that the message was deleted due to moderation action.
6. User B reports the message again. User A again receives a notification. From the dashboard, keep the message (i.e., do not take moderation action). The message in the chat returns to its original state (with the original text).
7. Verify the same steps again but using a how-to's chat. In this case, clicking the "view" button should take user A to the how-to in the how-to space.
8. Create a project outside of a gallery using User B's account. Add User C as a collaborator, start a chat, and send a message. Verify that the report button is not present. 
9. Create a new gallery using user A's account, and add user B as a creator. User B creates a project in this new gallery. User B adds user C as either a collaborator or commenter. User B sends a message; user C can see the report button, and clicking it sends a moderation notification to user A. 

## Checklist

- [x] If user A and user B are collaborators on a project that is in a private gallery, and user B is not a curator or creator on the gallery, they are not able to use the moderation action because they don't have read permission on the gallery (and thus can't tell that the project is in a gallery). It seems to be the same underlying issue as not being able to see that the project is in the gallery from the "share" dialog. 
    - If we want users A and B to be able to report messages for moderation, it should be a relatively straightforward fix (change the gallery prop for the ChatView component to be the gallery ID, which anyone with access to the project can see). Is this the desired behavior?
- [x] Is there anything from the [larger moderation proposal](https://github.com/wordplaydev/wordplay/issues/819#issuecomment-3640396603) that might need to be included?

### DIFF
--- a/src/edit/menu/PossibleEdits.test.ts
+++ b/src/edit/menu/PossibleEdits.test.ts
@@ -133,13 +133,6 @@ test.each([
         Replace,
         '-5',
     ],
-    [
-        'suggest addition on number with unit',
-        '5m',
-        (node) => node instanceof NumberLiteral,
-        Replace,
-        '5m + _•#m',
-    ],
 ])(
     '%s: %s',
     (


amyjko:
Thanks! I'll do a careful review of this on Saturday.

melissachen2000:
Thanks Amy! The page has been moved, and I double-checked the links from the gallery page, projects page, and notifications to make sure they're pointing to the new link. I've also filed the notifications feature as issue #938 .

PR #941, opened 2026-02-05
Speech to text

ZhiyuAlexZhang:
# Context

I am adding a new speech-to-text Input Stream called `Speech`, using the Web Speech API. 

This will enable voice input and appeal to those who can speak a language but are still in the process of learning how to read or write.

Wordplay users can configure the stream through:

1. `reset`: Clears accumulated speech when set to true
2. `language`: BCP-47 language code (e.g., en-US), defaults to en-US
3. `limit`: Max number of words to keep, defaults to infinite

## Related issues

N/A, this is a new feature. With that said, the localization team will need to translate the Speech Stream descriptions.

## Verification

I tested on both Chrome and Safari, and both work as intended. The testing was primarily done in English, with some additional testing in Chinese, Japanese, Korean, and Spanish.

## Checklist

N/A, we are ready to ship! 💯

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ These notes are publicly posted in [production](https://wordplay.dev/updates), s
 
 ### Added
 
+- We added a new @Speech stream, which works in most browsers. If you allow the site to use your microphone, you can make projects that respond to speech! (#394).
 - We added a machine translation of a Hebrew locale (#411).
 - We added a machine translation of a Vietnamese locale (#985).
 - We made autocompletion of conversions smarter and less buggy.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "wordplay",
-    "version": "0.16.44",
+    "version": "0.17.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "wordplay",
-            "version": "0.16.44",
+            "version": "0.17.0",
             "bundleDependencies": [
                 "shared-types"
             ],
--- a/src/routes/updates/updates.json
+++ b/src/routes/updates/updates.json
@@ -1,6 +1,6 @@
 [
   {
-    "version": "0.16.45",
+    "version": "0.17.0",
     "date": "2026-03-03",
     "changes": {
       "added": [


src/input/Speech.ts
```diff
@@ -0,0 +1,460 @@
...
+        this.on = true;
+
+        // Check browser support
+        if (typeof window === 'undefined') return;
+
+        const SpeechRecognitionAPI =
+            window.SpeechRecognition || window.webkitSpeechRecognition;
+
+        if (!SpeechRecognitionAPI) {
+            console.warn('Speech recognition not supported in this browser.');
```
amyjko: If the API isn't supported, the stream should return an error value as its value, rather than failing silently. See `Webpage` for an example of how to handle this.

src/input/Speech.ts
```diff
@@ -0,0 +1,460 @@
...
+    // Handles Web Speech API errors
+    // This should cover *ALL* possible errors in the API docs:
+    // https://webaudio.github.io/web-speech-api/#speechreco-error
+    private handleError(error: string, message: string) {
+        // Handle specific error types with appropriate messages and retry logic
+        switch (error) {
+            case 'network':
+                // Network error - could be no internet or service unavailable
+                this.react(
+                    'Network error: Check your internet connection and try again.',
```
amyjko: All of these messages should be localized, rather than hard coded as English.

src/input/Speech.ts
```diff
@@ -0,0 +1,460 @@
...
+    // Retry logic
+    // Backoff strategy: delay = baseDelay * attemptnumber
+    // Attempt 1: 1000ms; Attempt 2: 2000ms; Attempt 3: 3000ms
+    // After [maxRetries] failures, it injects a final error message and stops the stream to avoid infinite loops
+    private attemptRetry(errorType: string) {
+        // Only retry for transient errors, and only up to max retries
+        if (this.retryCount < this.maxRetries && this.on) {
+            this.retryCount++;
+            const delay = this.retryDelay * this.retryCount; // Exponential backoff
+            console.log(
```
amyjko: Remove this log before we finalize this PR.

src/input/Speech.ts
```diff
@@ -0,0 +1,460 @@
...
+                    try {
+                        this.recognition?.start();
+                    } catch (e) {
+                        console.error('Failed to restart recognition:', e);
+                    }
+                }
+            }, delay);
+        } else if (this.retryCount >= this.maxRetries) {
+            // Max retries reached
+            this.react(
```
amyjko: This should be a localizes string value for the stream, not a console error.

amyjko (changes requested):
This is very close! I'm impressed how well you picked up most of the patterns necessary to build a stream; it's one of the more complex parts of the programming language implementation.

Here are the key things to address:

- [ ] There are several error handling cases in the `Speech` implementation that either fail silently, or fail by printing to the console. All of these should be localized error values that become a new string value. See `Webpage` for a pattern for this. Essentially, when there's an error, we want that error to show up as the stream's value on stage. (I think is is more graceful then displaying an exception, since the stream already results in a text value).
- [ ] I believe your branch is out of sync with main; there are several type errors in see `npm run check` related to the `ColorJS` package that are resolved in main, but aren't resolved here.
- [ ] There are a few places where the we should and shouldn't be using the current locale selected for the UI. See comments for detail.

I will machine translate the strings prior to merging, once the error messages all have appropriate localization strings.

src/input/Speech.ts
```diff
@@ -0,0 +1,457 @@
...
+                this.attemptRetry('unknown');
+                break;
+        }
+    }
+
+    // Retry logic
+    // Backoff strategy: delay = baseDelay * attemptnumber
+    // Attempt 1: 1000ms; Attempt 2: 2000ms; Attempt 3: 3000ms
+    // After [maxRetries] failures, it injects a localized error message and stops the stream
+    private attemptRetry(errorType: string) {
```
amyjko: This error type is unused. Is this a case where the error is being ignored, or is this not needed?

amyjko:
@ZhiyuAlexZhang This is pretty close, just a couple of minor issues. Want to resolve this and we can ship?

ZhiyuAlexZhang:
Hi @amyjko! So sorry for not being able to make it to last week's meeting (my study session beforehand ran way over time past 7pm), and I just fixed both issues:
1. Moved the retry counter reset from `onstart` to `onresult`, and updated the auto restart logic in `onend`
2. Also cleaned up the unused `errorType` parameter in `attemptRetry`

Looking forward to catching up next Wednesday!

Issue #825, opened 2025-08-26
Community-created documentation


melissachen2000:
### What are you trying to do that you can't?

In a co-design session with 8th grade students and their teacher about sharing work in Wordplay, students mentioned a variety of obstacles related to:

1. Supporting students in developing projects for certain interests (e.g., game development, graphics). Currently, Wordplay has limited functionality for making games, such as platformers, which many students in the group were interested in making.
2. Learning how to implement something they had in mind. Students critique that the tutorial is long and hard to search within, and want tutorials for more specific things they want to make or shorter tutorials for specific concepts.
3. Difficulty making connections between Wordplay concepts and syntax with other programming languages that learners may be familiar with from prior classes.

### What is your idea?

Session participants suggested creating a community-created and community-curated documentation, like an internal wiki. This addresses the above by:
1. Allowing students to share snippets of code of code with each other that can be copy-pasted or drag-and-dropped into their editor. Snippets can have two goals: (1) to share a small bit of code that demonstrates a functionality, or (2) longer bits of code that enable a particular functionality, similar to how Python has modules that can be imported to add additional functionality (e.g. pyplot for data visualization). (Or, if having modules or libraries is something that would be possible in Wordplay and meets the design goals, especially those around transparency, that would also address the second goal)
2. Allowing students to create their own tutorials to teach others about how to implement something. Each tutorial can be a page of the wiki. Students suggested having interactive code sandboxes where you can modify the code, video tutorials (which will need additional design to support accessibility), and ways to walk through or annotate code. 
3. Allowing students to help each other make connections between other programming languages' syntax and concepts, for example, by indicating on an internal wiki page about `Phrase` that this is similar to `strings` in other languages. This may also help to improve searchability, where students can search for Wordplay concepts using terms that they are personally more familiar with (e.g., search for the concept of and/or syntax for `Phrase` using the search term `string`).

Participants illustrated the following features of this community wiki: 
1. Built-in question asking and answering, where students and teachers can ask questions and have them answered, with the answers being documented within the wiki. Relatedly, there should be a way to survey students and teachers for what documentation is needed, and, separately, for students and teachers to be notified when a new article related to one of their questions is added. 
2. A way to "chat with the documentation" by querying a chatbot that will point you to the section of documentation that may be useful. Participants noted that this might be particularly helpful for supporting students in making connections between Wordplay and other programming languages, and may be particularly useful "if you don't know what you're looking for"
[...]

PR #943, opened 2026-02-07
Adding user-written translation capability to how-tos

melissachen2000:
# Context

Currently, user-written how-tos can only be written and viewed in one language. This poses a problem if a user wants to write versions of a how-to in multiple languages. This PR sets up the user interface so that users can write how-tos in multiple languages.

The implementation uses markup strings to store translations. Each how-to is stored as a markup Doc with a locale tag. For how-tos made prior to this PR, the how-to text is just stored as a regular string. Since I accidentally hardcoded the locales on these how-tos to be `en-US`, I included some logic that detects if a string is not a markup string, and attaches the `en-US` locale to it. 

## Related issues

-   Related Issue #825 
-   Closes #

## Verification

1. Set up Wordplay to use more than one language using the language dialog in the bottom right.
2. Click the button to create a new how-to. Check that all of the languages that are added to Wordplay for the user are shown in the dropdown. 
3. Select a language with the dropdown. Write a title and some text. Repeat for a second language.
4. Post the how-to. In the database emulator, check that both the title and text are Wordplay markup strings, where the title and text are wrapped in ¶ symbols with corresponding language tags (e.g., `¶hello¶/en-US¶hola¶/es-MX`). 
5. Remove all of the languages but one. Open the how-to that was just created and click the edit button. Check that the two languages that that how-to was written in are still options in the drop-down. If the language that remains is not one of those two languages, check that the interface language is also in that list. 

To test that previously-written how-tos can still be read:
1. Start on a different branch than this one. Create a how to.
2. Switch to this branch. Refresh the page. Click the how-to preview. The text should render correctly. 
3. Click on the edit button. The language should be set to en-US.

## Checklist

- [x] Currently, `MarkupHTMLView` doesn't handle translations. Based on our conversation in Slack, it sounds like this is not implemented yet?
- [x] **Verifying**: I assumes 2-3 letter language code, 2 letter region code in my regex. Is this a correct assumption?
- [x] **Hoping to get help on:** Multilingual placeholder titles. A previous PR defined a placeholder title at `l.ui.howto.editor.titlePlaceholder`. We use this placeholder value if the title is left empty ([here is the implementation](https://github.com/wordplaydev/wordplay/blob/243ed90b6abb919f9da3b0ab0225ed3d01fe5ffb/src/routes/gallery/%5Bgalleryid%5D/howto/HowToForm.svelte#L212-L213)). I think having a placeholder title is probably still useful, but I don't know how to access the placeholder title for a particular locale. What I would like to be able to do is:
    - Have a list of locales that are used (defined as at least one of the text boxes in that locale is non-empty) (done)
    - For the list of locales that are used, get a list of locales that have empty titles (can do)
    - For the list of locales with empty titles, set their titles to be the placeholder value (don't know how to do)
- [x] **Hoping to get help on:** There is an announcer that announces whenever a how-to is moved in the preview. It uses the how-to's title. The current implementation gets the title in the user's preferred locale, if a title has been written in that locale, and an empty string otherwise. This feels like unideal behavior, because if a title exists, it should be shown. What should be the fallback?

### DIFF
--- a/src/locale/en-US.json
+++ b/src/locale/en-US.json
@@ -1,9 +1,7 @@
 {
     "$schema": "../../static/schemas/LocaleText.json",
     "language": "en",
-    "regions": [
-        "US"
-    ],
+    "regions": ["US"],
     "wordplay": "Wordplay",
     "term": {
         "bind": "bind",
@@ -1684,10 +1682,7 @@
                 "Some use the keyboard (/ctrl+9/ for \\⊤\\ and /ctrl+0/ for \\⊥\\). Some use the character search at the bottom of the editor. Or you can drag us from here.",
                 "Check out our @FunctionDefinition below. They're very logical."
             ],
-            "name": [
-                "⊤⊥",
-                "Boolean"
-            ],
+            "name": ["⊤⊥", "Boolean"],
             "function": {
                 "and": {
                     "doc": [
@@ -1697,10 +1692,7 @@
                         "\\⊥ & ⊤\\",
                         "\\⊥ & ⊥\\"
                     ],
-                    "names": [
-                        "&",
-                        "and"
-                    ],
+                    "names": ["&", "and"],
                     "inputs": [
                         {
                             "doc": "The other @Boolean to check. If the first is \\⊥\\, it doesn't matter what this is, the function will evaluate to \\⊥\\.",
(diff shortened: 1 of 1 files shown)


amyjko (changes requested):
This looks good overall. There are a few locale conflicts to resolve though; I'm not sure which keys you intended to be final. Can you review and resolve and then this is ready for merge?

amyjko (approved):

PR #1047, opened 2026-04-10
Update '教' header translation to '教育'

wd041216-bit:
# Context

‘教’ does not sound fluent when it appears in the form of single character, especially when other headers all appear in the form of double characters, so I update it to '教育'

## Related issues

-   Chinese locale Improvement #1042 
-   open #

## Verification

I verified through my 21 years experience of speaking Chinese

## Checklist

It is not a draft pr.


amyjko (changes requested):
Thanks for the PR! One minor change: when you see `$~`, that means the text was machine translated. Now that you've reviewed it, and you aren't a machine, remove that annotation, so we know it's human translated.

wd041216-bit (commented):
I have deleted the part covered by '$~', which I have reviewed

amyjko (commented):
Looks good, thanks!

wd041216-bit:
You are welcome

PR #1063, opened 2026-04-27
Defer construction of non-essential sub-databases.

amyjko:
The landing page only needs Settings, Locales, and the auth user, but the
Database singleton was eagerly constructing ProjectsDatabase, GalleryDatabase,
ChatDatabase, CharactersDatabase, HowToDatabase, and CreatorDatabase at module
import time. ProjectsDatabase.hydrate() opened IndexedDB on every page load and
the auth listener fanned syncUser() out to all six sub-databases on every
sign-in, even when the visited page never read any of them.

Make those six sub-databases lazy: they are now constructed on first property
access via getters on Database, and the top-level Projects / Galleries / Chats /
CharactersDB / HowTos / Creators exports are Proxies that defer to those
getters so existing import shapes still work. updateUser, login, and clean only
notify sub-databases that have already been instantiated; if a sub-database is
constructed after auth has fired, the getter calls syncUser itself so it
catches up to the current user.

Settings and Locales remain eager because the root layout needs them
synchronously for theming, fonts, and language direction.

### DIFF
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -49,23 +49,65 @@ export class Database {
     /** The database of loaded locales and settings. Encapsuled to avoid cluttering this central interface to persistence and caches. */
     readonly Locales: LocalesDatabase;
 
+    /** Lazily-instantiated sub-databases. Only constructed on first access so the
+     * landing page (and other surfaces that don't need them) avoid the IndexedDB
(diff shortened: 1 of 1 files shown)


amyjko:
This is very broken. Closing.

PR #1066, opened 2026-04-28
Allow how-tos to be shared publicly

### LINKED ISSUE #1052: Make how-to public without making gallery public

We want to be able to make a single how-to public without its associated gallery public. This will be particularly useful for sharing how-tos with class visitors so they can access the tutorial without being able to see the projects in the gallery.

### What is your idea?

I would suggest adding an option to the "Collaborate" panel of the how-to that allows the user to make it public. The explanatory text can say something like: "By making the how-to public to all Wordplay users, it can be viewed by anyone with the link." 

A caveat: I don't think it'll show up in the Guide at `/guide` since that looks at the galleries that the user has access to.

A consideration: with the current implementation, the person who views the how-to via a link would not be able to use the social features (commenting, attribution, chat), since users must be explicitly added as viewers in the HowToDatabase to have access to those social features. I think this is fine, and possibly even a good idea to reduce possible moderation needs. 

I will plan to work on this, starting this week! Filing an issue for tracking and for feedback on the design idea. Thanks!

melissachen2000:
# Context

Currently, to be able to view a how-to in the gallery, the user must have direct access to the gallery (curator or collaborator), indirect access to the gallery (via expanded permissions), or the gallery must be public. This is not effective in cases where a how-to should be accessible to all -- for example, when Adrienne wants to share a how-to she wrote with visitors from UW. 

This PR adds a "make public" feature to how-tos. In this case, anyone with the link can access the how-to, independent of other access settings. 

## Related issues

-   Related Issue #
-   Closes #1052 

## Verification

1. Create a how-to, do not make it public. Check that accessing the how-to via that link in an incognito window _does not_ grant access.
2. Make the how-to public by accessing the "share" panel in editing mode. Check that accessing the how-to via that link in an incognito window _does_ grant access via the incognito window.


melissachen2000:
Not sure why this test is failing, I will look into it later this week!

amyjko:
That test is a bit flaky; it's not you. I was going to work on making it less flaky this weekend. We won't let it block this PR.

amyjko (approved):
This PR looks good to me. I will merge, despite the flaky test failing.

Issue #1082, opened 2026-05-03
Public how-tos are not public


melissachen2000:
### What happened?

I am able to replicate this issue in production but not while testing locally with an emulator. 

1. Create a how-to; either make it public before publishing it the first time or edit it to make it public -- both are broken, as is making an existing how-to public
2. Copy the link to the how-to and paste the link into an incognito window (not logged in to Wordplay). The how-to should show up in the top left corner of the canvas, but it does not.

### Anything else?

No clue what's happening, but working on debugging! I can't seem to replicate the bug when testing locally, so I'm creating the issue for logging and help seeking. 

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

Windows

melissachen2000:
#1086 does not resolve the issue -- I'm continuing to investigate!

melissachen2000:
Resolved by d7796d47ae25560ce953f70d3eff7e1f1bcd6abf

PR #1086, opened 2026-05-03
Public how to bugfix

melissachen2000:
# Context

Currently, how-tos that were created prior to #1066 then changed to be public after that change was merged in are not viewable by the public. What appears to be the issue is that when the how-to is updated via the interface, the version number for the schema was not updated. 

## Related issues

-   Related Issue #1082 
-   Closes #

Not setting this as "closing" #1082 since I think we may need to double check that this does fix the issue in production. 

## Verification

When testing locally:
1. Check out commit f10333067ebde4f473e20af6760a2ea554c46970  -- this is a commit from before the public how-to feature was published. 
2. Create a how-to. This mimics a how-to that was created before that PR was merged in.
3. Change it back to this branch (e.g., with `git switch -`). This mimics PR #1066 being merged in but adds in our bugfix.
4. Edit the how-to and change its setting to "public"
5. Copy the how-to link and paste it into an incognito browser. Verify that the how-to is visible in that incognito window. 

I'm not 100% sure that this will fix the issue in production, but I have a good reason to believe that it is the issue because any how-to that used the version 1 schema but had the `isPublic` key set to true would have had it overwritten to `isPublic=false` by the upgrader. I think what would need to happen is that any how-to that was set to public would need to be updated again after this PR is merged in, which should set version to 2 and isPublic to the user's setting.

### DIFF
--- a/src/db/howtos/HowToDatabase.svelte.ts
+++ b/src/db/howtos/HowToDatabase.svelte.ts
@@ -109,7 +109,7 @@ const HowToSchemaV2 = HowToSchemaV1.extend({
     isPublic: z.boolean(),
 });
 
-const HowToSchemaLatestVersion = 2;
+export const HowToSchemaLatestVersion = 2;
 const HowToSchema = HowToSchemaV2;
 
 export type HowToDocument = z.infer<typeof HowToSchema>;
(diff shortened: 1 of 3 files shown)


amyjko (approved):
This looks sensible to me.

PR #1094, opened 2026-05-06
Fix how-tos not rendering if two how-tos are posted without refreshing the how-to space

### LINKED ISSUE #1083: Posting two how-tos without moving the camera leads to only one being rendered
### What happened?

1. In a how-to space
2. Create a how-to
3. Without moving the camera, create a second how-to. Only the second how-to shows up in the canvas and the drop down.
4. Move the camera a bit. Create another how-to. This third how-to will appear in the canvas and drop down.
5. Check the database (emulator) -- all three how-tos exist in the database, but only the second and third will render in the frontend. 

### Anything else?

The how-tos exist in the database, but for some reason the database query from the frontend only returns some of them. I don't know why this is happening, but I am starting to investigate. I checked the coordinates in the database, and the coordinates shouldn't be overlapping either, so it's not a case of two previews being rendered on top of each other.

Chrome

Windows

melissachen2000:
# Context

If a user posts two how-tos in a row without moving the camera or refreshing the page, only the second how-to is rendered and the first is not. It turns out that this is because the locally cached version of the gallery and the cloud version fall out of sync; the how-to ID for the first how-to is added, but then overwritten by the second how-to's ID in the local copy, so even though both how-tos exist in the cloud database, the first how-to doesn't exist in the cached local copy. This PR fixes this by reloading the gallery each time the HowToForm component is rendered, not just when the component is mounted. 

## Related issues

-   Related Issue #
-   Closes #1083 

## Verification

1. Create a new how-to space
2. Create a how-to (titled: test1)
3. Without moving the camera or refreshing the page, create a how-to (titled: test2)
4. Pan the camera to verify that both test1 and test2 exist in the how-to space

### DIFF
--- a/src/db/howtos/HowToDatabase.svelte.ts
+++ b/src/db/howtos/HowToDatabase.svelte.ts
@@ -536,16 +536,8 @@ export class HowToDatabase {
     }
 
     async getHowTos(howToIds: string[]): Promise<HowTo[]> {
-        let howTos: HowTo[] = [];
-
-        howToIds.forEach(async (howToId) => {
-            const howTo = await this.getHowTo(howToId);
-            if (howTo) {
-                howTos.push(howTo);
-            }
-        });
-
-        return Promise.resolve(howTos);
+        const results = await Promise.all(howToIds.map((id) => this.getHowTo(id)));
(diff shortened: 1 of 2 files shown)


amyjko (approved):
Looks good to me.

PR #1102, opened 2026-05-07
Fix how-to in Guide panel resetting scroll position when typing in Project editor

### LINKED ISSUE #1091: Guide pane scrolls to the start of the how-to after typing code
### What happened?

As reported by students, the guide pane in the project view scrolls to the top of the how-to whenever they type code into the project, which is annoying when they are following a how-to. 

### Anything else?

~I have not been able to replicate this on Chrome or Edge.~

I was able to replicate a slightly different version which possibly has a similar root cause, details in #1102 

Microsoft Edge

Windows

melissachen2000:
# Context

Students reported that the guide panel in the project viewer would scroll back up to the top of a long how-to whenever they typed. When I tried to replicate the issue, my experience was:

1. When I opened the how-to, the how-to would automatically scroll so that the caret positioned at the start of the last Example in the how-to would be just offscreen below the visible range of the how-to. 

image

2. I would scroll to a different part of the how-to, then type something into the editor, and the how-to would scroll back to the original view from when I first opened the how-to

It appears that the issue is that Wordplay will always scroll to where the Caret is positioned, which might be explaining why it was scrolling to the position of the last Caret in the guide. Disclosure: I used Claude Code to get help with identifying the cause of the bug. The solution it offered was unnecessarily complicated so I re-wrote it myself -- it checks whether the Caret is in an editor, and if so, scrolls; otherwise, it does not scroll. 

## Related issues

-   Related Issue #
-   Closes #1091 

## Verification

1. Create a long how-to with multiple examples (I used Adrienne's map how-to for debugging) in a gallery
2. Create a project in the gallery. Open the project editor for that project
3. Open the Guide pane in the project viewer. Verify that the how-to exists.
4. Open that how-to in the Guide pane. Verify that when the how-to is opened, the beginning of the how-to is seen.
5. Scroll to an arbitrary point within the Guide, and take note of where it scrolled to.
6. Type some Wordplay code into the project. Wait until the project re-compiles. Verify that the how-to did not scroll away from the point noted in step 5.
7. Type a lot more code into the project editor, until the source code is longer than can fit in the editor window. Verify that the editor will scroll down to show the last line of code in the source (that is, this change does not impact scrolling in the editor). 

## Checklist

- [ ] Editing is still laggy when I am testing locally, in a way that the production version of Wordplay does not lag. However, it is laggy both with and without the changes I made in this PR. This suggests to me that the lag is not due to this PR, but I wanted to note this in case it is introducing lag again. 
- [ ] I am worried that this may break something else in Wordplay

### DIFF
--- a/src/components/editor/caret/CaretView.svelte
+++ b/src/components/editor/caret/CaretView.svelte
@@ -89,17 +89,17 @@
 </script>
 
 <script lang="ts">
-    import { measureTokenSegment } from '@components/editor/highlights/measureTokenSegment';
-    import MenuTrigger from '@components/editor/menu/MenuTrigger.svelte';
-    import { getEditor, getEvaluation } from '@components/project/Contexts';
-    import { animationDuration, locales } from '@db/Database';
     import Caret from '@edit/caret/Caret';
     import type { LocaleTextAccessor } from '@locale/Locales';
     import Node from '@nodes/Node';
     import Token from '@nodes/Token';
     import { TAB_TEXT } from '@parser/Spaces';
-    import UnicodeString from '@unicode/UnicodeString';
     import { tick, untrack } from 'svelte';
+    import { animationDuration, locales } from '@db/Database';
+    import UnicodeString from '@unicode/UnicodeString';
+    import { getEditor, getEvaluation } from '@components/project/Contexts';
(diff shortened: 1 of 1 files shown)


amyjko (approved):
I've tested this and it is sound.

Btw, I've also noticed the local-only lag. I suspect it's related to when the web inspector is open.

PR #1111, opened 2026-05-11
Re-ordering how-tos and adding filter

### LINKED ISSUE #1087: New ordering for user-created how-tos in the project view panel

See #1085 -- as more how-tos are made available to students, it will become challenging to navigate within the guide panel in the project. 

### What is your idea?

We suggest adding a filter button to the top of the how-tos in the project panel. By default, the panel only shows how-tos that are in the project's gallery or are bookmarked from any gallery. There will be the option to allow the user to see all how-tos they have access to. 

We suggest the following ordering for how-tos in the project panel, building off of #1085 : 

1. Bookmarked how-tos from any gallery are at the top. There will be a 🔖 added to indicate that these are bookmarked
2. How-tos from that project's gallery are next
3. If the filter is set to "all how-tos", then all remaining how-tos are shown here

Then for the how-tos when viewed at `/guide`
1. Bookmarked how-tos are at the top, with a 🔖 symbol
2. How-tos from all galleries are next

melissachen2000:
# Context

Users who are creating how-tos for each other may not be in the same gallery (example case: one class section may be making how-tos for another, or one class may be making how-tos for next year's class). So, the current implementation that only shows the how-tos for the gallery that the project is in, if the project is in a gallery, does not address this use case well. This PR implements a new filter button and ordering for how-tos, available as follows:

_If the project is in a gallery and the user is viewing the Guide via the project editor:_
- Users will be shown Mode that lets them see all how-tos that they have access to or only the how-tos in that gallery and any that they have bookmarked
- The how-tos will be in the following order: (1) bookmarked how-tos from any gallery, (2) how-tos from the current gallery, (3) (if in "all" mode) how-tos from all other galleries

_If the project is not in a gallery and the user is viewing the Guide via the project editor OR the user is viewing the Guide via `/guide`:_
- Users will see all how-tos from all of the galleries they have access to
- The how-tos will be in the following order: (1) bookmarked how-tos, (2) all other how-tos

Bookmarked how-tos are displayed with a 🔖 symbol.

I wrote this PR with minimal debugging support from Claude Code for three small bugs (of which it could not fix the second). 

## Related issues

-   Related Issue #
-   Closes #1087 

## Verification

1. Create two galleries, gallery A and gallery B. Make one how-to in gallery A and two how-tos in gallery B, with distinct titles so it is easy to know which gallery each how-to is in. Bookmark one of the how-tos in gallery B. 
2. Open `/guide`. Verify that all three how-tos are visible and the Mode button for filtering is not visible. Verify that the bookmarked how-to is shown with a 🔖 . Verify that the bookmarked how-to is first. 
3. Create a project. Do not add it to a gallery yet. Open the project viewer and the Guide panel. Verify the same thing as in step 2. 
4. Add the project to gallery A. Verify that the filtering Mode button is now visible. 
5. Set the filtering button to "all," then verify the same thing as in step 2. Additionally, verify that the how-to from gallery A is shown second and the non-bookmarked how-to is shown third. 
6. Set the filtering button to "gallery + bookmarks only," then verify that: (1) the how-to from gallery A is visible, (2) the bookmarked how-to from gallery B is visible, (3) the non-bookmarked how-to from gallery B is not visible.
7. Set the filtering button back to "all" then verify per step 5. 
8. Remove the project from the gallery and refresh the page. Verify per step 3.


amyjko (approved):
This looks great, thanks Melissa.

PR #1121, opened 2026-05-13
Fix incorrect Chinese localization for 'auto' in settings panel

### LINKED ISSUE #1115: Incorrect localization for the setting panel
### What happened?

Image

Image

the option 'auto' was translated into '汽车', which means a car in Chinese. It should be translated into '默认' or ‘自动'.

### Anything else?

_No response_

Safari

macOS

wd041216-bit:
## Summary

Fixes #1115

The option 'auto' in the settings panel was incorrectly translated as '汽车' (car) in Chinese (zh-CN). This PR changes it to '自动' (automatic), which is the correct translation.

## Changes

- `static/locales/zh-CN/zh-CN.json`:
  - layout label: `"$~汽车"` → `"$~自动"`
  - theme (dark) label: `"$~汽车"` → `"$~自动"`

## Test plan

- [x] Verified the two occurrences of `汽车` were replaced with `自动`
- [ ] Review by native Chinese speaker (optional)


static/locales/zh-CN/zh-CN.json
```diff
@@ -5022,7 +5022,7 @@
                     "layout": {
                         "label": "布局",
                         "labels": [
-                            "$~汽车",
+                            "~自动",
```
amyjko: Please remove the `~` annotation. `$~` means that the text is machine translated, and it is no longer machine translated. Leaving just the `~` has no meaning, and will just be rendered in the editor.

amyjko (changes requested):
Please remove the out of date annotations.

wd041216-bit:
Fixed in 6f4c228. This removes the stale machine-translation markers from the two zh-CN "auto" labels so they now render as plain 自动.

static/locales/zh-CN/zh-CN.json
```diff
@@ -5063,7 +5063,7 @@
                     },
                     "dark": {
                         "label": "主题",
-                        "labels": ["$~光", "$~黑暗的", "$~汽车"],
+                        "labels": ["$~光", "$~黑暗的", "自动"],
```
amyjko: Since you're looking at this line, can you also fix the other labels? If they are correct, remove the machine translation tags.

amyjko (changes requested):

static/locales/zh-CN/zh-CN.json
wd041216-bit: OK, I have re translated the whole panel

amyjko (approved):

PR #1126, opened 2026-05-15
Add show less toggle for galleries

### LINKED ISSUE #1097: No “show less” option in Galleries section after loading more results.

In the Galleries section, clicking “More” shows more entries, but there is no way to collapse them back without refreshing the page.

Image

### What is your idea?

Add a “Show Less” option in the Galleries section so users can return to the previous view after clicking “More.” This would make it easier to navigate and manage long lists of entries in the future.

thomassinn24:
## Context

Adds a “Show Less” toggle to the Galleries section.

When expanded, the existing “More” button is replaced with “Show Less,” which resets the gallery view back to the initial 5 entries.

Closes #1097

## Verification

Tested locally by:
- Clicking “More” to expand galleries
- Clicking “Show Less” to confirm it resets to 5 entries
- Verified behavior in browser on desktop

## Checklist

- [x] Show More expands gallery
- [x] Show Less resets to initial 5 entries
- [x] No console errors

### DIFF
--- a/src/locale/en-US.json
+++ b/src/locale/en-US.json
@@ -5990,6 +5990,10 @@
                     "more": {
                         "label": "more...",
                         "tip": "load more galleries"
+                    },
+                    "less": {
+                        "label": "show less",
+                        "tip": "collapse to fewer galleries"
                     }
                 },
                 "error": {
--- a/src/routes/[[locale]]/galleries/+page.svelte
+++ b/src/routes/[[locale]]/galleries/+page.svelte
@@ -58,6 +58,8 @@
         nextBatch();
     });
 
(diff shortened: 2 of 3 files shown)


amyjko:
Hi @thomassinn24! Just a process note: you didn't get design approval for this before building it. The issue is still labeled `needs design`. The design you've built here doesn't quite make sense; the current page design is paginated for an arbitrary number of galleries, so it's not clear why a "show less" button makes sense. (e.g., if there were 1,000 galleries, what would "show less" do, and what would be the point of that?) Please discuss the design in the issue, iterate with me, get design approval, and then we can build whatever design we settle on.

PR #1131, opened 2026-05-17
Redefining expanded gallery definition

### LINKED ISSUE #1088: Redefine "expanded" how-to viewing permissions

Currently, if a how-to space for a gallery is set to "expanded" permission, then any user who is a creator or curator on ANY gallery that the curator curates can see the how-tos for that gallery. While that can be useful in some scenarios, in many cases it doesn't make sense for every creator and curator to be able to see that gallery's how-tos. 

### What is your idea?

Allow the curator to specify which gallery or galleries should have expanded access to the how-to space. Since the `viewers` key in the HowTo schema tracks which galleries grant viewer access, we should be able to achieve this functionality. Letting the curator specify which galleries have access to this gallery's how-tos may also help with the learnability of this feature.

melissachen2000:
# Context

The current definition of expanded galleries is hard to understand and too broad. Regarding the broadness problem, currently, if a gallery is set to "expanded" visibility mode, then all curators and creators in **all** of a curator's other galleries will be able to see the how-tos in this gallery. It would be more useful if the curator is able to specify which other galleries get access to this gallery's how-tos. For example, then a teacher could add Section A's gallery to Section B's gallery and vice versa, then the students can have access to each others' how-tos without having access to the projects.

Hopefully, by allowing curators to explicitly specify which other galleries get access to this one helps with the understandability of this feature, too?

## Related issues

-   Related Issue #
-   Closes #1088 

## Verification

1. Create two users, User 1 and User 2
2. From User 1's account, create two new galleries, Gallery A and Gallery B
3. Add User 2 as a collaborator on Gallery B
4. Check that User 2 does not have access to Gallery A's gallery or how-to space
5. Change Gallery A's visibility from limited to expanded. Check that User 2 now has access to Gallery A's how-to space
6. Remove User 2 from Gallery B. Check that User 2 no longer has access to Gallery A's how-to space.
7. Add User 2 to Gallery B again. Check that User 2 has access to Gallery A's how-to space.
8. Set Gallery A to limited. Check that User 2 no longer has access to Gallery A's how-to space.
9. Set Gallery A back to expanded. Check that Gallery B is in the list of galleries granted expanded access. Check that User 2 has access to Gallery A's how-to space.
10. Delete Gallery B. Verify that User 2 does not have access to Gallery A's how-to space.


amyjko (approved):
This is a nice tweak and much easier to reason about! I've done a gentle pass on the logic, and it seems sound to me.

PR #1057, opened 2026-04-23
1053 Animated Name Examples

ellipsis7449:
# Context

I have developed my Animated Name examples for Adrienne's classroom. It is one example, with two versions. Both of these are written primarily in English, but one involves my name written in English and the other has it written in Hindi. The project is linked below:

[Animated Name](https://wordplay.dev/project/7c642d23-bffe-41f6-8d6a-6e69f2d9ca75)

I am adding this project to the Visualizations gallery, since it's showing a visualization of text.

## Related issues

-   Related Issue #1053 
-   Closes #1053 

## Verification

Testing on MacOS Chrome and Opera GX (Chromium)
- Did try navigating the interface with only keyboard; although this pull request does not involve any changes to the UI and thus would not worsen current accessibility
- I have added language tags with every change in language (for Hindi example) to address multilingual accessibility

Issue #1053, opened 2026-04-21
Curate: Animated Name examples for Adrienne's classroom


ellipsis7449:
### What are you trying to do that you can't?

Since Adrienne and her students are working on animating their names with things that represent their identity, I wanted to make two examples using my name in both English and Hindi in the spirit of the project.

### What is your idea?

Here are my plans for the two projects:

**English name:**
- A: written in a fancy font, kind of ink-like to represent my passion for writing. Maybe rotating to somewhat emuluate me writing (I don't think I'd be able to animate a pen writing the letter with the knowledge I have of Wordplay)
- d: rotating on the y-axis and shifting between blue and green to represent the Earth, since I share my name with the Hindu goddess of the earth
- i: a purple quaver / lone eighth note jumping up and down, to represent my love for music and my favorite color
- t + i: two eighth notes beamed together and jiggling from side to side, as a continuation of my love for music and just a creative way to express it

**Hindi name:**
- अ: fading in and out like I tend to do with conversation, since I like observing more when in group settings
- दि: flashing purple because it's my favorite color
- ती: same animation as the 'd' in the English name, for similar reasons as that letter

(I have kind of already developed these two projects in Wordplay, but I just wanted to follow the workflow at least a bit and get feedback / ideas on if I should add more + what I should add. @amyjko I'm just gonna mention you right off the bat to be assigned to this.) Thanks!

amyjko:
These sound like great additions to the build in galleries! 

For your PR, you can create project files in `src/static/examples`, and then decide how to categorize them in `src/examples/examples.ts`. I'm not entirely happy with this current approach to built in examples, but it's what we have for now. Think about how you want to categorize these (including if we should have a new gallery for animated text like this).

PR #1139, opened 2026-05-21
1053 fixed animated name with multilingual islocale function

ellipsis7449:
# Context

Fixed the Animated Name Example that originally had two source files, one with each language. To make the multilingual nature of the example more apparent, I've used the islocale() function to include a Hindi version alongside the English version in the same file.

-   Related Issue #1053 
-   Related Issue #1057

## Verification

While I'm not sure how to make different output show up for each locale, I have managed for both outputs to show up simultaneously on the stage, which I think is good enough and makes it pretty clear that there's two versions.


ellipsis7449:
Okay, pretty sure my changes have been made for real and it works the way I intend it to now :)

amyjko (approved):

PR #1171, opened 2026-06-01
How to loading bug

melissachen2000:
# Context

Adrienne reporting lagging when navigating the how-to space as well as previews taking a long time to load.

**Previews**
- Problem: although a preview cache is being maintained, it is cached on the project ID, which is ephemeral (new null project created to render the preview). 
- Solution: cache how-to previews on a "cache ID" instead, which is the how-to's ID

**Navigation lag**
- Problem: the entire canvas was re-rendering each time it was dragged
- Solution: use CSS transforms, which apparently is a lot faster?

- Problem: we updated the canvas camera position or the preview tile's position for every mouse move instead of every paint (which has a lower frequency)
- Solution: only update the how-to's or camera's position when painting is done

- Problem: the canvas lags the most when approaching a new how-to that has not been rendered yet
- Solution: pre-load how-tos that are within half a canvas of the current view

While I had the hypothesis that the issue was that the canvas was being re-rendered every time it was dragged, Claude identified the problems and suggested each fix when prompted. I verified the changes.

## Related issues

-   Related Issue #
-   Closes #

## Verification

I added code to each of the pre-generated how-tos so that examples would be computed. Panning and moving behavior is intact. I don't think I'm seeing the same slowdowns as Adrienne is, but I can confirm that panning is faster and there is no longer a long pause before a new how-to is rendered on the screen.

### DIFF
--- a/scripts/seed.ts
+++ b/scripts/seed.ts
@@ -471,7 +471,7 @@ async function seedCreatorHowTos(): Promise<void> {
             guidingQuestions: HOWTO_GUIDING_QUESTIONS,
             text: HOWTO_GUIDING_QUESTIONS.map(
                 (_, qi) =>
-                    `¶Example answer ${qi + 1} for "${title}". (Seeded content — edit me!)¶/en-US`,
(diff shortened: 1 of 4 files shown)


amyjko (approved):
Reviewed, this all seems sensible and like it will resolve the canvas issues.

PR #1185, opened 2026-06-12
Feat: Fetching description from program

### LINKED ISSUE #897: Add project descriptions to project previews

Learn what a game is about.

### What is your idea?

Adding descriptions to what each gallery project does, in addition to the symbol previews.

tiaL-ops:
# Context

The issue was that the built-in projects at https://wordplay.dev/gallery/Games should contain some description. That is fetched from the first line `program ¶ ... ¶`.

To add the feature, I used regex to parse the first inline description, and segmented it using `Intl.Segmenter` to the language it is built in.

I also added a pencil so that the component is able to be used in programs, and the users and the owner of the program can edit it.

## Related issues

* Related Issue #897
* Closes #897

## Verification

I ran `npm run dev` and went to each built-in project at `http://localhost:5173/es-MX/project/` to see if it matches the description, and if there is not one, it is handled gracefully.

Ran `npm check` and `npm test` to make sure I did not break anyone else's code.

## Checklist 

* [x] Regex implementation
* [x] Intl Segmentation
* [x] Render view in project
* [x] Pencil to edit
* [x] Edit tested

## Review feedback:
* [x] Addressed review feedback. The implementation now uses the existing AST/documentation node APIs instead of regex-based parsing and full program rewrites.


amyjko:
Thanks for the initial draft! To log in locally, use `npm run emu`. The emulator creates seed accounts and logs in as `creator` automatically, so you can create a local project and test.

src/components/app/ProjectPreview.svelte
```diff
@@ -131,17 +130,157 @@
...
+    // A project's description is the main source file's leading documentation.
+    // It can be multilingual: a sequence of ¶…¶/lang blocks at the top of the
+    //
+    // We parse all leading blocks, then display the  one that is segmented 
+
+
+    type DescriptionDoc = { text: string; lang: string | null };
+
+    /** Matches one leading documentation block: ¶text¶ with an optional /lang tag. */
+    const DOC_BLOCK_REGEX = /^\s*¶([^¶]*)¶(?:\/([a-zA-Z-]+))?/;
```
amyjko: Documentation can be multilingual, so only matching on Latin characters will only work for languages that use Latin scripts. Instead, parse the program and get the Program node's `docs` field, which has the multilingual documentation text.

Additionally, Docs are a markup syntax, so we can't just pull arbitrary text. Instead, we should take the `markup` field of the first `Doc` in the array, and use `MarkupHTMLView` to render the markup in the preview. You'll have to test whether the segmenter works with our markup syntax; it may get confused by our various formatting symbols.

src/components/app/ProjectPreview.svelte
```diff
@@ -131,17 +130,157 @@
...
+        }
+        return { docs, rest };
+    }
+
+    let descriptionDocs = $derived(
+        parseLeadingDocs(project.getMain().getCode().toString()).docs,
+    );
+
+    /** Choose the doc that best matches the viewer's languages. */
+    let descriptionDoc = $derived.by(() => {
```
amyjko: This code is redundant with the functionality already built into `Docs`. Use its methods to choose the best locale to show.

src/components/app/ProjectPreview.svelte
```diff
@@ -131,17 +130,157 @@
...
+        // block, then whatever came first.
+        const primary = project.getPrimaryLanguage();
+        return (
+            docs.find((doc) => doc.lang === primary) ??
+            docs.find((doc) => doc.lang === null) ??
+            docs[0]
+        );
+    });
+
+    let description = $derived.by(() => {
```
amyjko: I'm not sure supporting editing on the preview is a good idea: it seems like it could easily break the initial Docs in the program text. Since it's only getting the first sentence, it would have to ensure that edits are carefully re-inserted into the original Docs markup. It looks like the save functionality below just overwrites the whole thing, invisibly erasing whatever other sentences were there.

src/components/app/ProjectPreview.svelte
```diff
@@ -131,17 +130,157 @@
...
+            .map(
+                (doc) =>
+                    `¶${doc.text}¶${doc.lang !== null ? `/${doc.lang}` : ''}`,
+            )
+            .join('\n');
+        const body = rest.replace(/^\n+/, '');
+        const newCode =
+            prefix.length > 0 ? `${prefix}\n${body}` : body;
+
+        Projects.reviseProject(project.withSource(main, main.withCode(newCode)));
```
amyjko: This is a very risky way to modify the program, since it rewrites the whole thing, rather than just the Docs portion of the `Program` node. Rather than doing a whole replacement of the entire source text, just update the `docs` node of the Program node specifically.

amyjko (changes requested):
This is a nice start! The general functionality is moving in the right direction. The key things to fix are to reuse the existing functionality in the platform to read and edit nodes. The current implementation is a bit hacky, using regexes and re-writing the whole program, rather than doing more targeted AST reads and writes to the program's documentation node.

src/components/app/ProjectPreview.svelte
tiaL-ops: Thank you for the comment. Yes, it is cleaner to directly use the AST. 
In this new version I use : 
`project.getMain().expression.docs.docs[0] ?? nul,`
That is, I get the description from the project source code, dig into the expression, and get the first item of the docs.

With the current state, I tested with English, French, and Korean, where it did not need Intl.Segmenter but instead directly took the first component.

src/components/app/ProjectPreview.svelte
tiaL-ops: Yes, editing in preview is pretty complicated.
I do propose two ways to allow editing:
1- Editing is only allowed in the preview; users would need to directly modify the program.
2- Allow editing, but link it to the program so that any modification rewrites and updates the  docs nodes of program.

In this current version, I use option (1); I removed the pencil. However, please do not hesitate to let me know if there is another way to handle editing.

src/components/app/ProjectPreview.svelte
tiaL-ops: That makes sense. This is dependent on how editing will be updated, hence I wanted to confirm whether the inline editing should stay or not; then I  would be happy to work on this

amyjko (approved):
I made a few revisions:

* Earlier versions were partially but incorrectly using Intl.Segmenter; I added full support.
* I encapsulated first sentence extraction to Markup.ts, so others can reuse it, rather than it living in ProjectPreview. I added tests for that.
* I did a bit of layout refinement because large chunks of text were vertically centered with the buttons, which created weird vertical space.

PR #1205, opened 2026-06-25
Begin Nepali localization

AJT14-cmd:
# Context

Begins the Nepali (`ne-NP`) localization. The `term` and `token`
sections are manually translated. Other entries retain their `$?`
English fallback text and still require translation and review.

## Related issues

-   Related Issue # 1204

## Verification

- Ran Prettier on `ne-NP.json`
- Ran the locale verifier
- All 2,870 unit tests pass

## Checklist

- [x] Translate `term` and `token` in ne-NP.json
- [ ] Translate remaining locale sections
- [ ] Translate tutorial
- [ ] Add emoji annotations
- [ ] Translate how-to content


amyjko:
Thank you for this, @AJT14-cmd, and I'm sorry it sat open so long.

**Nepali is on the site.** It shipped in c946044f7, with the interface, the tutorials, the how-tos and the emoji names. You started that, and it's live for Nepali readers now.

Because it landed on `main` separately, this branch can no longer be merged, and I want to be precise about why rather than just closing it:

- `main`'s `ne-NP.json` now has about 5,500 strings; this branch has about 3,700, and most of them are still `$?` — the marker meaning "not translated yet". Merging would replace Nepali text with English placeholders.
- The branch is also built on an older shape of the files. It still has `ui.collaborate.role.collaborators`, for example, which `main` has since renamed to `role.collaborate`.

**Your hand-written translations are not lost.** 103 of them are in #1402, with you as co-author. Several fix places where the machine translation had plainly failed — `main` has been shipping these in the token names that appear all through the editor and the guide:

| Path | What `main` shows today | What you wrote |
| --- | --- | --- |
| `token.Function` | `$~function` | प्रकार्य |
| `token.SetOpen` | `$~set/map खुला` | समुच्चय खोल्ने |
| `token.EvalOpen` | `$~evaluation खुला` | मूल्याङ्कन खोल्ने |
| `token.Access` | `$~property पहुँच` | गुणमा पहुँच |

Two things I couldn't carry over, so you know where they went:

- **50 `term.*` strings.** That section no longer exists in the English source, so there's nowhere in the current files to put them.
- **`token.URL` and `ui.localize.emotion`.** These had no marker, but the text was still the English. `npm run locales` flagged both when I first tried to include them — an unmarked English string would ship as though it were Nepali.

**If you'd like to keep going, the most useful thing now is review, not translation.** Almost all 5,500 Nepali strings are marked `$~`, meaning a machine wrote them and no person has checked them — and as the table above shows, the machine left English in the middle of Nepali sentences in places. You can edit any string from inside the site at https://wordplay.dev/localize, which opens a pull request for you; there's no need to clone anything or work with `git` unless you want to. Your `token.*` fixes are exactly the kind of correction that's needed.

Closing in favour of #1402. Thank you again — and if you'd rather reopen this and bring the branch up to date yourself, say so and I'll leave it to you.

Issue #1073, opened 2026-04-29
Missing API usage limits and/or throttling


ETBRamos:
### What happened?

The absence of API throttling or rate limiting for users creates the opportunity for malicious actors to easily utilize bots on one or more devices to rapidly request translations. Due to the google translation API billing by character, an attack this vulnerability could lead to massively inflated costs of operation (I believe this is called API gateway flooding for DoW attacks). If an API quota limit is set on the API side, this would still take up the entire quota for other users of the platform.

### Anything else?

My initial thought is implementation of IP and session-based rate limiting as well as global throttling, as a failsafe in cases where multiple devices are utilized in an attack. I'd like to work on this issue, but it isn't my area of expertise, and I'd like to hear others' ideas and critiques of mine prior to planning implementation if possible. A captcha system would also be great to use, for example, if the system detects bot-like behavior, it could be required the next time the translation is requested, but I'm not yet sure of how the implementation would work in this particular case. I think some sort of indicators, such as a specific error message for a limit being hit, should also be provided to users so that they understand what has happened. 

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

Windows

amyjko:
Thanks for reporting @ETBRamos! One thing to do first is to evaluate the current attack vector. What is possible in the current architecture? That will inform what remedies need to be in place.

amyjko:
Also, @ETBRamos, do you want to be assigned?

ETBRamos:
Hi, yes, I would like to be assigned to this issue. A full evaluation of the range of possible current attacks would require access to the cloud API backend to access the current configurations there , but from what I can see on the frontend, the API appears to be just called directly without checking anything from the user. 

Additionally, Here's what I've found so far regarding what can be done for security on Google Cloud backend as you'd requested:

Most of the inbuilt security for Google's cloud APIs, including the Cloud Translation API seems to be locked behind the Google Cloud Armor service in the google cloud console, and while this has essentially all of the capabilities we’ve been looking for, it is not a free service. This a brief overview of what could be done with it:

Traffic limiting via the throttle strategy in can be enforced via IP:
--enforce-on-key-configs user-ip

Or multiple keys such as:
   --enforce-on-key-configs="HTTP-PATH,HTTP-COOKIE=site_id"

These would return a ‘429 Too Many Requests’ 

[...]

Issue #938, opened 2026-02-01
Notifications for reporter and message author during moderation process


melissachen2000:
### What are you trying to do that you can't?

See issue #819 for design for moderation and PR #926 for its implementation. The primary feature missing from the larger proposal for moderation is that the user who wrote the message that was reported and the user who reported the message do not get notifications when a message is reported or when moderation action is taken. 

### What is your idea?

1. User whose message is reported should be notified when the message is reported
2. User who reported the message should be notified when moderation action is taken
3. User whose message is reported should be notified when moderation action is taken

Issue #430, opened 2024-03-29
Support multiple language tags on a single text literal or doc


amyjko:
## What's the problem?

Text can be multilingual, e.g., `"Holá gentleman!"`, but Wordplay currently only allows text literals, names, and documentation to have a single language and region. We support multiple alternatives in different languages for separate monolingual translations `"Holá"/es"Hello"/en`, but not multilingual content.

## What's the design idea?

Allow text literals, names, and documentation to have multiple language codes and regions on them. For example:

`"Holá gentleman!"/en-US|es-MX|es-SP`

That can't be the exact syntax, but it conveys the point: some text is multilingual.

## Design

See [approved design](https://github.com/wordplaydev/wordplay/issues/430#issuecomment-2738036778).

diana24p:
I would like to work on this needs design issue please. @amyjko

amyjko:
This is a good one! Design things to do:

* Figure out the syntax
* Figure out how multilingual text should be evaluated. (We currently choose the text string that has the closest matching current output locale; what to do with mixed language content)?

diana24p:
Right now Wordplay picks one version of a text based on the user’s preferred language setting. Instead of Wordplay choosing only one language, we need to support mixed-language text where both languages are stored and recognized together. 

If the user prefers English, Wordplay should prioritize English words while keeping the Spanish parts intact. 
Example: If a user sets their language preference to English and runs into the text "Hola gentleman!", the entire phrase should still be displayed but Wordplay should recognize and prioritize the English word "gentleman" when processing or translating the text. Maybe bolding or highlighting prioritized words could be helpful? Even though Wordplay would prioritize the user’s preferred language, it should still recognize and process all languages in a phrase.

Example of what we want to do: 

"Hola gentleman!" --> Marked as Spanish (es) + English (en)

Three ways to mark multilingual text.

Bracket notation: ["Hola gentleman!": "en|es"] 
- text is inside square brackets []
- languages are listed after a colon (:), separated by |

JSON style key-value: { "text": "Hola gentleman!", "languages": ["en", "es"] }
- a key/value pair
- good for storage

Inline tagging (angle brackets): "Hola gentleman!<en|es>"
- Uses < > to tag languages

amyjko:
Thanks for the brainstorms @diana24p! Just a few clarifications:

* This issue isn't about the visual presentation of multilingual text, it's about the encoding of language tags in the syntax of text literals.
[...]

PR #1214, opened 2026-07-02
feat: Chat translation 

### LINKED ISSUE #771: Translate chat messages

When teachers and students are chatting with each other, if they use different languages, they may not understand each other.

### What is your idea?

We could add a translate button that would translate the chat text into the current user's locale. This could be optionally saved in the chat, so that we don't redundantly ask for a translation.

tiaL-ops:
# Context

In this pull request, I am implementing a feature to translate chat messages.

## Related Issues

- Related Issue #771
- Closes #771

## Verification

- Ran `npm run dev` and `npm run emu`, then opened Collaborate and Chat to verify that the dropdown renders correctly.
- Checked `localhost:4000` and the Firestore emulator to verify that chat messages are sent and that the schema is updated.
- run npm run check:now, npm test,  , all test passed 
- npm run locales : the x label are : 4 unwritten ("$?") string(s) would fall back to English. Run "npm run locales-translate" to fill them: ui.collaborate.translate.label, ui.collaborate.translate.language, ui.collaborate.translate.error, ui.collaborate.translate.messageError

## Checklist

- [x] Updated the Message V3 schema to include a `language` field (the language the message is written in).
- [x] Verified the changes in the Firebase Emulator.
- [x] Added a dropdown at the top of the chat box before sending, using `getTranslatableLocales()`.
- [x] Added another dropdown at the top to translate messages.
- [x] Added a function that uses `translateMarkupText` with the LLM as the raw translator.
- [x]  Have a fallback design when chat not working
- [x] Verify translation by npm run locale


amyjko (changes requested):
This is a solid start! This is going in the right direction; you're using the right translation infrastructure, the message schema looks right to me. The core architecture looks solid.

Here are a few things to work on:

* All the user-facing text you've added is hard-coded into English. You'll need to update UIText.ts, generate the new locale schema with `npm run schemas`, run `npm run locales-fix` to add them to the locales, then move the strings to `en-US.json`, then update the UI to use the various localized text components to render them. There are many examples across the platform for uses of `<LocalizedText>`.

* Make sure to run `npx prettier` to keep formatting consistent.

* Translation is has a large number of sequential network round-trips. `translateMessages` loops over every message and awaits one `translateMarkupText` call each. For a long chat that's going to be slow. The `RawTranslator` interface takes an array of texts and returns them 1:1 — so you can collect all messages sharing a source language, translate them in one call, and map the results back. Definitely add that performance improvement.

* The issue (#771) explicitly suggests optionally saving the translation so we don't re-request it. Right now translations live in transient $state and a message sent after you turn translation on won't get translated until you toggle the dropdown again. I think we should persist translations in the message schema rather than re-translating them, to keep costs down. They should be invalidated if the chat message is edited.

* In the translation block you render both msg.language (source) above the divider and translations[msg.id].language (target) below. Two bare language codes stacked around a rule reads a little cryptically. Consider a clearer "from X → Y" display, or dropping the redundant source tag.

* Remember to `run npm run check:now`, `npm test`, and `npm run locales` to verify everything.

amyjko (changes requested):
This is good progress! It's getting closer. The batching is done right
(grouped by source language, one call per group, groups running concurrently), the
translations are cached and reused, the error handling degrades per-batch instead of
failing the whole chat, and the strings are properly localized with the right format tags.
The architecture is sound overall. 

I did a very careful review today, so the list is long, but everything should be within reach. Most of what's below is about making it actually run, and about some edge cases in chat that translation touches.

## Why translation "is still in English"

`ChatView.svelte` imports `functions` directly from `@db/firebase` and bails when it's
falsy. But `functions` is `undefined` until something calls `getFunctionsInstance()`
(`src/db/firebase.ts:109`) — that's what lazily loads the Functions SDK and wires the
emulator. Nothing in the chat path calls it, so you're hitting the error branch almost
every time. It probably looked intermittent, because `CreatorDatabase` happens to call
`getFunctionsInstance()` when it fetches missing creators, so whether translation works
is a race with an unrelated fetch.

Every other call site in the codebase does `await getFunctionsInstance()` —
`Translate.svelte:68`, `LocaleChooser.svelte:117`, `accountExists.ts:13`, `Join.svelte:45`,
`Login.svelte:84`, `localize/+page.svelte:802`. None of them import `functions` directly.
Do the same and I think your last checkbox unblocks itself.

While you're there, drop the top-level `import { httpsCallable } from 'firebase/functions'`
and import it dynamically like `src/db/projects/translate.ts` does — the static import
defeats the lazy load and pulls the Functions SDK into every page that renders a chat, which is a lot of extra code for the initial bundle.

## Rebase on current `main` before anything else

The branch is 58 commits behind and GitHub reports conflicts, and I think that's causing
most of the noise in this diff:

- `static/schemas/LocaleText.json` shows +22,520 / −21,806, but if I normalize both sides through prettier, only **36 lines** actually differ — the rest is 4-space → 2-space reindentation. Same for `zh-TW` (1,478 lines, 20 real), `ja-JP`, `hi-IN`. Running `npx prettier --write` on every changed JSON, as CLAUDE.md asks, collapses this to a diff we can actually read.
- More seriously, some existing translations lost content. In `es-MX`, `Paragraph.doc` went  from `"\¶Párrafo 1.\n\nPárrafo 2.\n\nPárrafo 3.¶'tres párrafos'\"` to just `"\¶Párrafo 1."` — text deleted, and an unterminated `\…\` block left behind. Same in `Refine.doc`, `PropertyReference.doc`, `StructureDefinition.doc`. `bn-BD` has the same   truncation plus whole strings quietly re-translated with markup markers dropped (`/সীমাবদ্ধ/` came back as the untranslated `/restrict/`) and the register changed. There's smaller damage in `mr-IN`, `ne-NP`, `zh-CN`, `ja-JP`, `hi-IN`, `vi-VN`, `id-ID`.

None of that is your fault — you ran the tooling as documented, but on a branch predating the fixes. `main` has since made `npm run locales` read-only (it reports drift instead of repairing it) and fixed the array-splitting rules that caused those truncations. So: merge `main`, revert all the locale and schema changes, re-run `npm run schemas` and `npm run locales-fix`, then prettier. The only locale changes left should be your four new keys and their `$?` placeholders.

## Correctness

- **Translation bypasses moderation.** `translateMessages` skips only `msg.text === null`, so `pending` and `removed` messages still get translated — and the `{#if translations[msg.id]}` block is a sibling of `<div class="what">`, so it renders outside the moderation guard just above it. That means turning on translation shows reported and removed content to people who shouldn't see it, and writes a translated copy of it to the database. Please mirror the same  `msg.moderation === undefined || msg.moderation === 'approved'` condition.

- **Deleted messages keep their translations.** `deleteMessage` nulls out `text` but leaves `translations` intact, so a deleted message's content lives on in translated form. This is the invalidation point from my last review — there's no edit path in chat, so delete and moderate are the equivalents.

- **Translations aren't counted against the chat size cap.** `Chat`'s constructor trims messages against `MAX_CHAT_MESSAGES_BYTES` (128 KB) using only `message.text?.length`. Cached translations are invisible to that estimate, so a chat translated into a few languages can be several times bigger than the trimmer thinks, and drift toward Firestore's hard 1 MB document limit. After that, sending a message starts failing.

- **The write side isn't batched, even though the reads are.** After batching the LLM calls, the caching loop calls `saveMessageTranslation` once per message, and each one runs a full transaction that reads and rewrites the entire chat document. Translating a 50-message chat is 50 serialized transactions on one doc. Do a single `saveMessageTranslations(chat, language, Map<id, text>)` that applies them all at once.

- **A Small race condition** `translations = next` is assigned after several awaits with no guard, so changing the dropdown twice quickly can let an older pass overwrite a newer one. Capture the target and bail if it changed.

## A design question

Right now translations only recompute when the dropdown changes, so a message that arrives while translation is on stays untranslated until you toggle it off and on. We might consider an explicit translate button instead of triggering on drop down change.

## Reuse

- The `RawTranslator` adapter in `ChatView.svelte` is a copy of the one in `src/db/projects/translate.ts`. Let's extract it once (something like `getFirebaseTranslator(functions): RawTranslator`) and use it from both.
- The ~130 lines of grouping and batching belong in `src/db/translateMarkup.ts`, not in a
view component. Tat module already owns this layer, is deliberately backend-agnostic, and has tests. A plural sibling of `translateMarkupText` would fit naturally, and would be unit-testable.
- For the language pickers: `Translatable` has ~161 languages, so both dropdowns are ~161 items in a narrow chat panel. `Translate.svelte` already solves this with `LocaleSearch` + `filterLocalesByQuery` and renders entries with `LocaleName` — I'd reuse those instead of the hand-rolled `languageOptions` / `languageName()`.
- Deduplicating the translatable locales down to bare language codes drops the region, which collapses `zh-CN` and `zh-TW` into `zh` — Simplified and Traditional become indistinguishable, and the same happens to `pt-PT` / `pt-BR`. Store the full locale string and compare with `localesAreEqual`, like `Translate.svelte` does.
- `MessageSchemaV3` and `V4` in one PR can just be a single V3. Also worth knowing: messages don't carry a `v` field, `MessageSchemaLatestVersion` isn't referenced anywhere, and the unknown-version union doesn't do anything since all the added fields are optional. I'd leave that machinery as you found it.
- `src/db/chats/ChatDatabase.test.ts` already tests `addMessage`, `reportMessage`,
  `moderateMessage`, and `deleteMessage` against a mocked Firestore. Tests for the language tag, the translation cache, and deletion clearing translations would be good.

## Localization and accessibility

- The two `role="status"` divs are live regions, and we route all dynamic announcements
  through `Announcer.svelte` via `getAnnouncer()` — component-local live regions cross-talk with the central one on real screen readers. 
- For the visible error, `Notice.svelte` is the component we use elsewhere (see `Translate.svelte`).
- The visible label and the `Options` `label` prop are the same string, so the control gets
  announced twice and has no distinct name. Pick one.
- A few string issues: the "off" option is a hard-coded `'—'` rather than a localized string; `error` and `messageError` are nearly identical, and `messageError` says "this chat" when it's about a single message; and the `→` between languages is hard-coded, which points the wrong way in RTL locales. A `Template<['from','to']>` would let translators control the wording and the direction — remember our template inputs are always named, never numbered.
- `field.language` is correctly tagged `[plain]`, but `field` is meant for input fields — it
  probably belongs next to `translate.label`.

## Housekeeping

Prettier hasn't run on the source files either: there's a stray blank line inside the
`@db/translateMarkup` import, and `ChatDatabase.svelte.ts` re-indents some pre-existing code it didn't otherwise touch.

## Suggested order

1. Merge `main`, revert the locale/schema churn, regenerate, run prettier.
2. Fix `getFunctionsInstance()` and verify translation actually runs against the emulator.
3. Moderation guard and deletion invalidation.
4. Extract the shared translator, move batching into `translateMarkup.ts`, add tests.
5. Localization and accessibility cleanup.
8. `npm run check:now`, `npm test`, `npm run locales` — and read the `locales` output for `x`-marked errors rather than trusting the exit code alone.

This is close! Ping me when you're ready for another pass.

tiaL-ops:
Thank you so much for all details! I will start working on it now and will let you know once its ready again

tiaL-ops:
<html><head></head><body><p>Here is an update on my work.</p><p>Regarding the <code inline="">LocaleText.json</code> conflicts, I merged with <code inline="">main</code>, regenerated the locales, and reran <code inline="">npx prettier --write</code>. Everything is now up to date, so I am not sure  which  way  should i do to "fix" the conflicts.</p><p>Also, when I run <code inline="">npm run locales</code>, aside from the expected missing translations that need Claude key, I see these delimiter mismatches:</p>

ar-SA | Paragraph.doc
bn-BD | Paragraph.doc
es-MX | Paragraph.doc, PropertyBind.doc, PropertyReference.doc, StructureDefinition.doc
id-ID | Paragraph.doc
ne-NP | Paragraph.doc, PropertyBind.doc, PropertyReference.doc, StructureDefinition.doc
pa-IN | Paragraph.doc
pt-PT | Paragraph.doc
vi-VN | Paragraph.doc
zh-CN | Paragraph.doc, PropertyBind.doc, PropertyReference.doc, StructureDefinition.doc

<p>I'm not sure whether I should fix these, since they already match the current codebase and there aren't any merge conflicts.</p></body></html>

amyjko:
Those conflicts mean that you changed something that was also changed in `main`. You should review them and see what you changed that is in conflict. Based on my brief skim, it looks like your PR is updating many locale string unrelated to this feature, so those need to be cleaned up.

tiaL-ops:
hi @amyjko ,  I jsut cleaned the unecessary change on json file, I think it is ready for a review now, let me know what you think!

amyjko:
Great, I will try to review Saturday morning.

amyjko (changes requested):
Thanks @tiaL-ops! This is good progress. I finished my review and have a bit more polish for you. Apologies for the volume; since there's real $ cost here, I want to be thorough.

## Defects
* `translations: undefined` is written into a Firestore payload, which is not permitted in Firestore. [ChatDatabase.svelte.ts](vscode-webview://09sthd709c0c6f933pt33inaj016qq9gsqjsjt5oeqkkho7edr9b/src/db/chats/ChatDatabase.svelte.ts) now adds translations: undefined to the transforms in reportMessage, moderateMessage (non-approved), and deleteMessage, and those objects go straight into tx.update(chatRef, { messages }). initializeFirestore in [firebase.ts](vscode-webview://09sthd709c0c6f933pt33inaj016qq9gsqjsjt5oeqkkho7edr9b/src/db/firebase.ts) doesn't set ignoreUndefinedProperties, so the SDK should throw Unsupported field value: undefined. The new unit test passes only because Firestore is mocked, and toMatchObject({translations: undefined}) is a near-vacuous assertion either way. The field has to be removed via deleteField() or by rebuilding the object without the key.
* A failed cache write is reported to the user as a translation failure. In translateMessages, translations = next happens before await Chats.saveMessageTranslations(...). If that write is rejected (e.g. a gallery moderator who isn't in participants — firestore.rules:154 only allows participants to update), the catch sets translateError = true, so the user sees translations and "There was an issue translating this chat."
* Cached translations are hidden behind the network call. Messages with a cached translation are collected into next but nothing is displayed until `translateMarkupTexts` resolves; if that throws, the cached ones are discarded too. Cached results should render immediately.
* Language codes and locale strings are mixed. messageLanguage defaults to $locales.getLanguages()[0] (a language code, "en") but the picker assigns localeToString(locale) ("en-US"). That means that the default is never shown as selected (`localesAreEqual` compares regions), messages are tagged inconsistently, the `msg.translations[target]` cache misses across the two forms, and `translateMarkupTexts` groups "en" and "en-US" as separate source languages. That's two LLM calls for one language, which isn't great for cost.
* Untagged messages are assumed to be in the viewer's UI language (msg.language ?? $locales.getLanguages()[0]). Every pre-existing message will be declared as, say, English to the translator regardless of what it's actually in.
* Messages already in the target language are still sent for translation. There's no `from === to` skip.
* Stale translating flag. The finally clears translating even on the discarded pass (if (target !== translateTo) return), so a superseded call turns off the spinner while the newer one is still running.
* Translation doesn't follow new messages. `translations` is a one-shot snapshot taken when a language is picked; messages arriving afterward render untranslated with no indication, even though the chat is visibly "in translate mode."

## Data model / cost
* Cached translations count against MAX_CHAT_MESSAGES_BYTES (the trim loop in the Chat constructor was updated to include them). Caching into N languages multiplies chat size by roughly N+1 against a 128KB cap, so translating starts silently dropping the oldest messages from everyone's view. It's worth deciding whether translations should be part of that budget at all, or capped/evicted separately.
* There's no ceiling on the number of cached languages per message, and saveMessageTranslations writes the whole message array regardless of size. Enough languages and the chat doc approaches Firestore's hard 1MB limit, at which point all chat writes fail.
* Any participant's translation choice mutates the shared document for everyone, and each new target language is a full-history LLM pass with no debounce or rate limit. Clicking through several languages is several full-chat translations.
* Consent/privacy: this sends other people's chat messages to Claude, and persists translations of their words into a doc all participants read. Project translation is the author's own content; this isn't. We should mention this in the policy document.
* Reporting a message wipes its cached translations, so a single report forces everyone to re-pay for translation of that message.

## Accessibility & conventions
* Announcements use `getMultilingualText`. It should use `getPrimaryPlainText` for `Announcer` text — the joined string reads every chosen locale back to back.
* Both announcements are constant strings. `messageError` says "There was an issue translating this message" without naming which message, and in the queued lane consecutive duplicates dedupe. This means that new announcements won't be read.
* I'm not sure we should add two new announcer kinds (`chat-translation-error`, `chat-message-errors`). Consider whether the existing banner/interrupt lane covers this rather than adding chat-specific kinds.
* The "off" button's accessible name is an em dash. translate.off is "—" and is used as both label and tip. Those should be more descriptive.
* The option lists are an unlabeled ad-hoc combobox — no aria-expanded, no listbox/option semantics, no arrow-key navigation, no result-count feedback. Selection is conveyed only by a border color, with no aria-pressed/aria-current; the selected option is also active={false}, which drops it from the tab order without explanation. Those are WCAG bugs that need to be fixed.
* Two `LocaleSearch` fields on one screen share the same default placeholder and description, so both inputs have the identical accessible name and neither says what it selects.
* `translate.language` ("message language") is documented as the ARIA label for the message-language selector but is actually passed as the per-option tip.
* The translate options reuse `l.ui.project.button.destination` as their tip — a project-translation string, which semantically wrong.
* `{:else}&mdash;` is the untranslated, which is non-descriptive empty-results state for both lists.
* Options only appear after typing. There's no way to browse available languages, and no visible indication of the current target once the query is cleared.

## Locale
* Unrelated translations were clobbered. [bn-BD.json](vscode-webview://09sthd709c0c6f933pt33inaj016qq9gsqjsjt5oeqkkho7edr9b/static/locales/bn-BD/bn-BD.json) and [zh-TW.json](vscode-webview://09sthd709c0c6f933pt33inaj016qq9gsqjsjt5oeqkkho7edr9b/static/locales/zh-TW/zh-TW.json) have rewritten ui.collaborate.tour.* strings that lose their /…/ and *…* markup and now contain an untranslated literal /restrict/. bn-BD also has a truncated [formatted] doc array element — "\\¶অনুচ্ছেদ ১।" — where the original was a three-paragraph example with a closing \. These need to be reverted.
* `off: "—"` is a string that can't be translated and doesn't say what it does; a real label ("Stop translating") would be better and would fix the button's accessible name too.

## Tests
* `translateMarkupTexts` — the new grouping/partial-failure logic, and the most testable thing in the PR — has no unit tests (grouping by source locale, per-group failure isolation, non-string entries, null result).
` No test for `Chat.withMessagesTranslations`.
* The trim test is placed inside `describe('ChatDatabase granular message operations')` above its own `beforeEach` and doesn't use `db`; it belongs with the Chat tests.

## Cleanup
* [LocaleSchema.ts](vscode-webview://09sthd709c0c6f933pt33inaj016qq9gsqjsjt5oeqkkho7edr9b/src/util/verify-locales/LocaleSchema.ts) and package-lock.json each have the trailing newline stripped as their only real change — both should be reverted.
* [ChatView.svelte](vscode-webview://09sthd709c0c6f933pt33inaj016qq9gsqjsjt5oeqkkho7edr9b/src/components/app/chat/ChatView.svelte) isn't prettier-formatted (verified against the PR head); the new markup blocks are hand-wrapped. Also a stray double blank line in the <style> block, and a trailing-whitespace comment line at the uniqueLanguageLocales declaration.
* `uniqueLanguageLocales` dedupes `getTranslatableLocales()`. It's worth confirming that source can actually produce duplicates rather than papering over it here.
* The `filterLocalesByQuery` change (adding `locale.language` and `localeToString(locale)`) is a reasonable improvement but silently changes the settings locale search too.

Once these are done, we should be good to translate and write a CHANGELOG.

tiaL-ops:
Thank you for the review, I will start updating the code

src/util/verify-locales/LocaleSchema.ts
tiaL-ops, line 1: I reverted/copy pasted mutliple times , I m not sure why it keeps moving the import place

tiaL-ops:
Hi @amyjko , I think I’ve made good progress on the work. I’d appreciate another review when you have time. I was thinking of updating the changelog once everything is finalized, or is it better to update it with every change instead?

amyjko (changes requested):
Thanks for your patience, I was on vacation and conferencing last week!

We're getting close. The architecture is basically there, there are just a few defects and polish to finish. 

## Defects

**1. `deleteField()` inside an array throws.** This is the same bug as the last round with the `translations: undefined`. `modifyChatMessage` rewrites the whole `messages` array, and `deleteField()` is now sitting inside an array element in `deleteMessage` and in `moderateMessage`'s non-approved path. I ran it against the real SDK:

```
Function updateDoc() called with invalid data.
deleteField() is not currently supported inside arrays (found in document chats/c1)
```

So deleting your own message and moderating a reported one both fail. The unit tests pass because `deleteField` is mocked to `{_op:'deleteField'}` and then asserted with `toEqual(deleteField())`, so the test encodes the broken behavior rather than catching it. The fix is easy: the transform already rebuilds the whole array, so just rebuild the object without the key, the way `trimChatTranslations` already does with `const { translations, ...rest } = message`. The test should assert the key is *absent* (`'translations' in msg === false`), which would have caught this.

**2. The two error notices render their placeholders literally.** `translate.error` is `Template<['to']>` and `translate.messageError` is `Template<['sender']>`, but both are passed to `<Notice text={...} />`, and `Notice` → `LocalizedText` never concretizes. Users see:

> Couldn't translate this chat to **$to**.
>
> A message from **$sender** couldn't be translated.

`Template<Names>` is `string & {...}`, so it's assignable to `LocaleTextAccessor` and TypeScript won't catch it. The announcer paths for these same two strings *do* concretize correctly, so it's just the visible ones. Either concretize and pass through `<Notice>`'s children snippet, or make these plain strings if the inputs aren't wanted on screen.

**3. Picking a language issues two full-chat translation passes, and throws one away.** Tracing it: `translateMessages` assigns `translateTo` (a `$state`) at the top; the content-key `$effect` reads `translateTo`; so the assignment invalidates the effect, which re-runs, finds `lastTranslationContentKey` still `''` (its own off-branch set it), and calls `translateMessages` again. That bumps `translateRequest`, so the first pass hits `if (request !== translateRequest) return` and discards results it already paid for — and it's the pass that would have cached them. The 300ms debounce doesn't help, because the duplicate comes from the effect, not from the user.

I traced this rather than executed it, so please double check that `getLLMTranslations` invocations in the emulator log for one language pick. There are likely two.

To fix this, make `queueTranslateMessages` do nothing but assign `translateTo` (debounced), and let the `$effect` be the single place that starts a pass. One trigger, no race, and the request counter goes back to guarding only genuine overlaps.

While you're there — the effect's content key deliberately omits `msg.translations`, which is the only thing preventing a save → snapshot → retranslate → save loop with unbounded LLM cost. That's completely invisible; add a comment saying so.

## Cost and data model

There are a few things I'm worried about with cost:

- **Translation eviction can oscillate.** When cached translations exceed `MAX_CHAT_TRANSLATIONS_BYTES`, the `Chat` constructor drops the oldest ones on read and `modifyChatMessages` drops them on write. Once a busy chat sits at the cap, the next pass re-translates what was evicted, saves it, which evicts others, and so on — paying for the same messages repeatedly. A per-language or per-message cap, or simply not re-translating messages older than the eviction frontier, would stop the churn.
- **Every new message costs a full-array rewrite per translating viewer.** With the live effect, each arriving message triggers one LLM call plus one (up to ~256KB) document rewrite for each viewer with translation on, all transacting on the same doc. In a classroom gallery with a lot of chat that's real contention. Can you think of a way to reduce that cost?
- `translateMarkupTexts` doesn't dedupe identical texts within a group, though `RawTranslator`'s own doc comment says it receives "the unique source strings." De-duplicate to save us tokens.

## Design

**Requiring a message language before you can send is a big UX change.** `messageLanguage` defaults to `undefined`, so opening any chat shows a red error `Notice` ("Tag your message with the language it was written in") and disables Send until you pick from a ~161-item dropdown. That's probably a bad UX choice for anyone who's not translating chats. `startChat` already defaults the chat's language to `localeToString($locales.getLocale())` — do the same for the message and keep the picker as an override. The red error styling for a state the user hasn't done anything wrong in is also too strong.

## Accessibility and localization

You should move to the platform's standard `<Options>` widget, rather than doing a custom `<select>`. It encodes accessibility features and ensures that it looks the same as other options. It also enforces localization.  As you do, a few considerations:

- **The label is still duplicated**, which is the specific thing I flagged last round. The translate bar renders a visible `<span>` "Translate messages to" *and* an `aria-label` with the same string *and* a placeholder `<option>` with the same string — so it reads "Translate messages to [Translate messages to ▾]" visually and announces the name twice. Use a real `<label for={id}>` (which also makes the label clickable) and give the placeholder option its own text.
- **The message-language select has no visible label at all** — its only visible text is the placeholder, which is the ARIA-label string `"message language"`, lowercase, written to be spoken rather than read. It needs a visible label and a distinct placeholder.
- **`ui.collaborate.translate.results` is dead.** It was added to `UITexts.ts` and shipped as `$?` to all 26 locales, but nothing references it (it was for the old search UI). Remove it before translation runs, so we don't pay to translate a string nothing uses.
- **Turning translation off is a separate button** that only appears once a language is chosen. An "off" option inside the select would be one control instead of two.
- The privacy paragraph is good and I'm glad it's there, but it sits immediately after "We don't store anything else," which reads as a contradiction. Moving it up next to the data-collection bullets would read better.

## Reuse

`getMultilingualLanguageLabel(locale)` in `LocaleText.ts` already is `getLocaleLanguages(locale).map(c => Languages[c]?.name ?? c).join(' + ')`. The PR reimplements that expression three times — twice in `ChatView`, once in `LocaleCombobox.optionText`. `optionText`'s region suffix is likewise `getLocaleRegionNames`. Deduplicate.

The `LocaleSearch` changes are now orphaned: chat doesn't use `LocaleSearch` anymore, so making `placeholder`/`description` required and then passing back the exact defaults that were removed is churn with no behavior change. The `filterLocalesByQuery` haystack change (adding `locale.language` and the full locale code) now only affects the settings locale search. Revert both, or split them out as their own small PR if you want the search improvement on its own merits.

`addChat`/`addChatToHowTo`/`addMessage` all take `language?` optionally but each has exactly one call site that always passes it — they can just be required.

## Housekeeping

- **`package-lock.json` and `LocaleSchema.ts` still have their trailing newline stripped** as their only real change. Revert both.
- **Prettier still hasn't been run**: `ChatDatabase.svelte.ts`, `ChatDatabase.test.ts`, `translateMarkup.test.ts`, `LocaleCombobox.svelte`, and `LocaleSchema.ts` all fail `--check`. (`ChatView.svelte` and `translateMarkup.ts` are clean now.)
- `translateMarkup.test.ts` has an `import` statement in the middle of the file; move it to the top.
- The `MessageSchemaV3` doc comments say "a Wordplay language code, e.g. `en`" and translations "keyed by target Wordplay language code (e.g. `es`)", but the values are full locale strings (`en-US`, `es-MX`). Fix the comments so the next reader doesn't reintroduce the code/locale mixing we just got rid of.
- The oversized-translations test is under `describe('Chat.withMessagesTranslations')` but exercises the constructor.
- `LanguageCode.test.ts` uses `await import('@locale/Locale')` inside the test — a static import at the top should work; if it doesn't, that's worth knowing why.

## Merge

The branch is 69 commits behind and the **only** conflict is `static/locales/es-MX/es-MX.json` (main's recent Spanish fixes). Take main's version and re-run `npm run locales-fix`. The locale diff is otherwise clean now — 13 lines per locale, exactly the new keys — which is a big improvement over last round.

## Suggested order

If it were me, I'd do:

1. Fix `deleteField()` (rebuild without the key) and change the test to assert absence.
2. Concretize the two error notices.
3. Collapse the double translation pass — `queueTranslateMessages` sets `translateTo` only; the effect owns starting passes. Comment why `translations` is absent from the content key.
4. Default `messageLanguage` to the viewer's locale.
5. Label fixes, remove the dead `results` key, "off" as a select option.
6. Reuse `getMultilingualLanguageLabel`; revert the `LocaleSearch`/`LocaleChooser` changes; revert the two newline-only diffs; run prettier.
7. Merge main (only es-MX conflicts), `npm run locales-fix`, prettier the JSON.
8. `npm run check:now`, `npm test`, `npm run locales`, then CHANGELOG.

amyjko:
I did some work on project translation and ended up creating some translation API caps as part of it, to prevent overuse and abuse (#1073). This feature will need to use this new backend. Here are some instructions on where this new functionality lives and what to change.

## 1. Delete your `src/db/getFirebaseTranslator.ts` — `main` has one now

Same path, same default export, same shape: `getFirebaseTranslator(functions) => RawTranslator`. It gained an optional second argument (`{ progress }`) you can ignore, so your `ChatView` call site is unchanged. In the rebase, your version of the file becomes a deletion.

You inherit for free:

- **chunking** — 25 strings / 2,000 characters per request
- a **300s callable timeout**, matching the function's own (the SDK default is 70s, which a long request can exceed)
- the caller's IANA **time zone**, which the daily budget resets on
- **partial-failure handling** — one failed chunk no longer loses the whole batch
- **budget-error classification** (see §6)

## 2. Never call `httpsCallable(functions, 'getLLMTranslations')` directly

The callable now rejects requests over `MAX_TEXTS_PER_CALL` (50) or `MAX_CHARACTERS_PER_CALL` (3,000 source characters) with `invalid-argument`. Going through the factory keeps you under both.

Also, `getTranslations` — the Google Translate fallback — has been deleted, along with the `@google-cloud/translate` dependency. Drop any reference to it.

## 3. `translateMarkupTexts` needs no change

It takes an injected `RawTranslator`, so it inherits everything above. Two behaviors to be aware of:

- `null` now means **every** chunk failed.
- A partly-failed group comes back as an array with `undefined` holes.

Your existing per-id `failed` handling covers the second case. Just confirm a `null` marks the whole group failed rather than being read as an empty result.

## 4. Translating now requires a signed-in caller

The callable rejects unauthenticated calls with `unauthenticated`. Chat already implies an account, but a session can expire while a view is open, so handle the rejection rather than assuming it can't happen.

## 5. The budget is shared, and chat is the riskiest consumer

Every creator gets 10,000 source characters per day (`DAILY_TRANSLATION_CHARACTERS`), enforced server-side. Project translation draws on the same pool.

Chat differs from the project dialog in a way that matters here: it translates on a 300ms debounce from an `$effect`, so it can spend a creator's whole day's budget without anyone pressing a button. Opening one long conversation shouldn't cost someone their ability to translate a project. Three things to add:

- **Gate the auto-translate effect** on `remaining() > 0` from `@db/translationBudget.svelte`.
- **Drop in `<TranslationMeter compact />`** (`@components/app/TranslationMeter.svelte`) so the spend is never invisible. It renders nothing when signed out and manages its own Firestore listener, so there's nothing to wire up.
- **Add a minimum-length guard**, so "ok" and "👍" don't each cost a request.

Your sidecar cache already avoids re-translating on reopen. Keep that.

## 6. Telling failures apart

The factory calls `noteTranslationRefusal` before returning, so `budget.refusal` is already set by the time you see a failure:

| `budget.refusal` | what to show |
| --- | --- |
| `'over-budget'` | the meter's reset time; don't retry |
| `'unauthenticated'` | prompt sign-in |
| `'failed'` | your existing generic error |

Call `resetTranslationRefusal()` when a new attempt starts.

The factory **stops asking** after an over-budget or unauthenticated refusal, so a long chat won't fire twenty doomed requests before giving up.

## 7. Rebase note

`src/db/projects/translate.ts` changed substantially on `main` — it gained a `rewrite` flag and plan/progress callbacks. If your branch touched that file, take `main`'s version wholesale.

---

Two other things I noticed, separately.

**The `chatTranslations` rule is `allow read, write: if request.auth != null`.** That lets any signed-in user read a private chat's translated messages, and write text that participants would render as another person's message translation. It should be scoped to the chat's participants the way `/chats/{chat}` is.

**There are two translation caches** — `MessageSchemaV3.translations` and the `chatTranslations` sidecar — and only the sidecar is ever written, so `withMessagesTranslations` has no production caller. Probably worth collapsing to one before this merges.

amyjko (changes requested):
Thanks for sticking with this! There's good progress here: `translateMarkupTexts` is a faithful plural of `translateMarkupText`, its grouping/dedup/partial-failure semantics are the right ones, and the seven tests you wrote for it cover the cases I'd have asked for. Extracting `getFirebaseTranslator` and reusing it from `translateProject` was the right instinct (and, as noted separately, `main` now has its own). So we definitely want to pull in that new work, as in my previous comment.

Below are some things we need to fix before this is ready.
---

## Blocking

### 1. Every new message re-translates the entire chat

This is the one I'd fix first, because it multiplies the cost problem rather than adding to it.

`translateMessages` builds its cache map from `msg.translations?.[target]` — but nothing ever writes `MessageSchemaV3.translations` (`withMessagesTranslations` has no production caller). So `cached` is always `undefined`, `next` starts empty, and then:

```ts
translations = next;   // wipes everything the sidecar subscription put there
```

The sidecar's `onSnapshot` results live **only** in `translations`, and this line clears them at the top of every pass. The pass runs on every content-key change — i.e. every new message, every moderation change, every delete. So a 50-message chat with translation on re-sends all 50 messages to the LLM when message 51 arrives.

It doesn't self-correct on reopen either: `onSnapshot` delivers asynchronously, so by the time the sidecar's first payload lands, `translateMessages` has already run synchronously, wiped `translations`, and queued every message. The comment on the subscription effect — "preventing a duplicate LLM call" — isn't what happens in practice.

The fix is to keep sidecar results in their own state that a pass reads from and never clobbers, and to consult it when building `toTranslate`.

### 2. Deleting or moderating a message leaves its translation in Firestore

`deleteMessage` and `moderateMessage` carefully strip `translations` from the message — but the translations that actually exist are in the `chatTranslations` sidecar, and nothing anywhere deletes a sidecar doc (I grepped: `ChatTranslationsCollection` appears only in `saveMessageTranslations` and `subscribeChatTranslations`). `forgetChat`'s comment says callers handle it; no caller does.

So the text of a deleted or moderation-removed message survives indefinitely as its translation — and under the current rule, any signed-in user can read it. The `isVisibleMessage` guard hides it in this UI, but the data is still there and still fetchable. Same question for chat deletion and account deletion.

### 3. `npm run locales` fails and prettier fails

- `npm run locales` exits 1 with 29 errors — 7–9 unwritten `"$?"` strings per locale (more than the 4 in the PR description). Needs `npm run locales-fix`. I will machine translate when the rest of the PR is ready.
- `npx prettier --check` fails on `ChatView.svelte` and `LocaleName.svelte`.

### 4. Hardcoded English in the UI

```ts
{ value: undefined, label: () => 'Choose a language' }
{ value: undefined, label: () => 'Use current language' }
```

These are user-visible strings that need locale keys. (`Option.label` accepts a plain `string`, so the `() =>` wrapper isn't buying anything either.)

### 5. The pickers' visible labels don't match their accessible names

Both selects pass `label={(l) => l.ui.collaborate.translate.language}` ("message language"), and `Options` turns that into `aria-label`. But each also has a visible `<label for>` reading something different — "Translate messages to" and "Message language". `aria-label` wins, so a screen reader announces a name that isn't the visible one (WCAG 2.5.3), and the two controls announce **identically**. The target picker should use `translate.label`, the compose picker `translate.messageLanguageLabel` — or drop the `aria-label` and let the `<label>` do the work.

---

## Should fix before merge

### Dead code and unused strings

- **`src/components/settings/LocaleCombobox.svelte`** — 131 new lines, imported by nothing. Please delete it (or use it; see the dropdown-size note below).
- **`ui.collaborate.error.untaggedMessage`** — added to `UITexts.ts`, `en-US.json`, and all 29 locales as `"$?"`, and never referenced in any component. That's a string 29 translators will be asked to review for a message that never renders.
- **`deleteField`** mock added to `ChatDatabase.test.ts`; `deleteField` isn't used anywhere in source.
- **`MessageSchemaV3.translations`**, `trimChatTranslations`, `messageTranslationsSize`, `evictedTranslationIDs`, and `withMessagesTranslations` are all machinery for the message-level cache that never runs. Collapsing to the sidecar (as I mentioned separately) removes all of it — including the eviction rule, which as written would permanently blacklist an evicted message from ever being translated again.

### Unrelated changes

Three files in this diff aren't about chat translation, and one of them has a bug:

- **`LocaleChooser.svelte`** now renders `<LocaleSearch id="locale-available-search">` **twice** — same `id`, and both bound to the same `query`. That's a duplicate-id axe violation and two fields that mirror each other. Also, the `placeholder`/`description` props being passed are exactly `LocaleSearch`'s own defaults, so they're no-ops.
- **`LocaleName.svelte`** — swapping the per-name `<span class="name">`/`<span class="join">` markup for a flat `getMultilingualLanguageLabel()` string changes how every locale name in settings renders, and drops the comment explaining the #430 behavior. It also leaves a stray whitespace-only line.
- **`LocaleSchema.ts`** — import reordering only.

I'd pull all three out; the `LocaleChooser` search field looks like it wants its own PR.

### `modifyChatMessage`'s type was widened unnecessarily

```ts
transform: (m: SerializedMessage) => Record<string, unknown>
```

This was `=> SerializedMessage`, and it looks like its type was widened only so the rest-destructure that drops `translations` would typecheck. It didn't need to be: `translations` is optional, so `Omit<SerializedMessage, 'translations'>` is already assignable to `SerializedMessage`. As it stands, `tx.update(chatRef, { messages })` now writes an unchecked `Record<string, unknown>[]` — the whole messages array lost its type. Please restore the narrow signature.

### `any`

```ts
label: (_l: any) => getMultilingualLanguageLabel(locale)
```

Four occurrences, and none are needed — `Option.label` takes `string | LocaleTextAccessor`, so this is just `label: getMultilingualLanguageLabel(locale)`.

### Visible text goes through `.toText()`

The three visible strings (`translate.error`, `translate.messageError`, `translate.direction`) are rendered as `$locales.concretize(...).toText()`. That flattens to a single primary-locale string, so it skips the multilingual echo and the markup rendering. `MarkupHTMLView` takes exactly this case:

```svelte
<Notice>
    <MarkupHTMLView markup={[(l) => l.ui.collaborate.translate.error, { to }]} />
</Notice>
```

### The translation isn't marked with its language

`<div class="translation">` carries no `lang` or `dir`. A screen reader reads the Spanish translation in the UI-locale voice, and an Arabic or Hebrew translation renders LTR inside an LTR chat. Both are one attribute each: `lang={translateTo}` plus the direction for that locale.

### Nothing is announced when translation succeeds

Picking a language produces an announcement only on failure. The `<Spinning />` is unlabeled, so it's silent too. A screen-reader user selects a language, waits, and hears nothing whether it worked or not — which per the announcer convention is indistinguishable from a broken feature. A completion announcement naming what varies (count and target, e.g. "translated 6 messages to Spanish") would cover it.

### The 674-option dropdowns

`getTranslatableLocales()` returns 674 locales. Two always-visible `<select>`s means ~1,350 `<option>` elements in every chat tile, and no way to filter — which is why `LocaleSearch` exists for the settings dialog. It also permanently costs two rows of vertical space in what is often a small tile. Worth considering: a much shorter list (the languages already present in the chat, plus the viewer's), or the search-filtered picker, or putting both behind a disclosure.

Related: a `ta-IN-LK-SG` viewer is the one supported locale *not* in that list, so `messageLanguage` initializes to a value with no matching `<option>` and the compose picker renders in a state the user never chose.

### The privacy paragraph doesn't match the implementation

Two mismatches:
- It says "the text of *other participants' messages*" — but the pass translates the viewer's own messages too (nothing filters on `msg.creator`). Either filter them, or reword.
- It says results are "stored in the chat document that all participants share" — they're in `chatTranslations`, which under the current rule is readable by every signed-in user, not just participants. Once the rule is scoped, this sentence becomes true; until then it understates what's shared.

---

## Smaller things

- `queueTranslateMessages` debounces the *assignment of the target*, not the translation pass, so a deliberate dropdown selection — and the "Stop translating" button — both sit for 300ms before anything happens. The debounce belongs on the pass.
- `messageLanguage` is `$state(localeToString($locales.getLocale()))`, evaluated once. If the viewer changes UI language with the chat open, their messages keep getting tagged with the old one.
- `catch (_) { translateError = true; }` discards the error entirely — worth a `console.error` so a failure is diagnosable.
- The chat schema bumps to v3 for one optional field, and `upgradeChat`'s `default` branch **throws** — so an already-open tab on the previous deploy breaks the moment a newer client upgrades a chat it's watching. Meanwhile `MessageSchema` gained two fields and its `MessageSchemaLatestVersion` constant was *deleted* rather than bumped. Those two decisions point in opposite directions; is the chat bump needed at all, given `language` is optional?
- `...(language !== undefined ? { language } : {})` in `addChat`/`addChatToHowTo` — `language` is typed `string`, not optional, so the conditional can't be false.
- `subscribeChatTranslations` ignores `!snap.exists()`, so a deleted sidecar never clears what's on screen.
- The sidecar doc has no size bound at all (the 128KB cap you added applies to the message-level cache that's never written), and Firestore's hard limit is 1MB per document.
- `chatTranslations` is a raw string literal rather than a `Domain` entry — worth a comment on why it's outside the five sync domains, since it's also outside the Dexie mirror and the backup scripts.
- `<hr class="divider">` is a semantic thematic break used as decoration; a CSS `border-top` on `.translation` says the same thing without adding a landmark to the reading order.
- The `//// msg.translations is deliberately excluded...` comment has four slashes and unindented continuation lines.
- Typo: "Show cached results right away then **aftere** network call".
- `active={chat !== undefined && newMessage.trim() !== ''}` was reflowed across two lines with no semantic change — that's part of the prettier failure.

---

## Design questions

1. Should a creator's own messages be translatable at all? Skipping them would cut cost, simplify the privacy story, and match what the policy paragraph already claims.
2. Is per-message language tagging worth the compose-time picker? The chat-level `language` plus a per-message default would get most of the benefit without a second dropdown in the composer.

tiaL-ops:
Thank you! I made some modifications. For a few parts, I have some questions.

### Design
As per our discussion a few months ago, a person might speak different languages, so having the option to pick a different language may be useful.

Should I go ahead and rule that out and make sure the default language is the user locale?

### About the locales

For locales, we do have some more than the 4 expected to translate:

Here is what each locale means:

* `label` — the translate-bar label and the Options label
* `off` — the "stop translating" button (tip and text)
* `error` — the pass-failed notice, and the announced error
* `messageError` — per-message failure, shown inline *and* announced (single-error case)
* `messageErrors` — the plural announced case (multiple failures)
* `direction` — the "from X to Y" tag under a translation
* `messageLanguageLabel` — the composer language picker label
* `choosePlaceholder` — the translate-to dropdown's "choose…" option
* `currentLanguagePlaceholder` — the message-language dropdown's default option
* `moreLanguages` / `fewerLanguages` — the search-expand toggle tips (both pickers)
* `translating` — the Spinning label during a pass
* `translated` — the announced success

Is there anything that you recommend removing then?

### Dead code

I used deleted files in the database in Chat.

**`LocaleSchema.ts`** — I tried reverting / `npx prettier`, but it still stays. It is weird. I don't know how the file is changing.

amyjko:
* They should be able to pick from multiple languages. But the default should be the primary selected locale.
* We want to make sure there's no hard-coded English, so if that's the complete list, then that is the right list. Before I approve, I will review the English, then machine translate it to all other languages.
* I'm not sure what you mean by `LocaleSchema.ts` "staying". Do you mean that it is being modified after you revert it? What's being modified about it?

tiaL-ops:
That makes sense, yesm the default locale was what I was going for.

Yes, that sounds good for the language. Let me know if there’s anything I can change.

I also just fixed some changes in the locale schema . Some imports were in the wrong place ( it was on my end), causing unnecessary changes to show up in my PR even when I hadn’t made them. I was struggling to remove them because they kept being pushed as changes, but I just fixed it.

amyjko:
I will review today! I may also do a bit of extra feature work on this: I just learned that Chrome has an API that will do translation for free on-device, and we might as well use that when available. Expect to see some commits from me. (And since this is very close, I will probably just finish the PR, since you've already gotten it to 95%).

amyjko (approved):
Thanks again for all of these contributions! The core of this is yours and most of it survived intact. `translateMarkupTexts` (batching by source language, deduplicating, isolating failure per group), the `lang/dir` props on `MarkupHTMLView`, never translating the reader's own messages, and the `LocaleSearch` disclosure instead of two 674-option dropdowns are all still here. Per-message language tagging turned out to be more important than it looked: browsers that translate on-device require an explicit source language and can't infer one.

One thing I did was catch the branch up to `main`, which had a bunch of chat related changes about moderation (#938) that rewrote the chat subsystem this builds on. Both branches claimed chat schema v3 for incompatible things, and main's is deployed with 115 production chats migrated — so language was added to v3 in place rather than as a v4. An optional field needs no upgrade, and a bump would break any tab still on the previous deploy, since upgradeChat's default branch throws. ChatView was rebuilt on main's version rather than line-merged, because the old one read msg.moderation in eight places and called reportMessage with an argument the callable now takes from the auth token.

I also fixed several defects that were preventing translation. The effect that schedules a pass cleared any pending timeout on its way past, and it re-runs for reasons unrelated to the conversation. A re-run inside the 300ms debounce cancelled the scheduled pass and returned early without setting another. Nothing threw or logged.

There was also a big security hole: the `chatTranslations` collection was readable and writable by any signed-in account, so translated text from any private conversation was exposed. It's now a subcollection under the chat, scoped to participants, with eviction moved server-side into report/moderate and a trigger for orphans.

For markup, both backends now have \code\ held out and @Concept links masked before translation. The prompt asked models to preserve them; they rename them anyway.

I added on-device translation via Chrome/Edge's built-in Translator, tried before our servers. It's free, private, and works offline.

I also made some interface changes from testing: the picker's first option is "Don't translate" (so "Stop translating" only appears while a pass runs), failures are dismissible, "nothing to translate" is an informational note rather than an error, and a translation now takes the body font size with the original smaller above it. We also persist language preference now.

tiaL-ops:
Awesome! Thank you,
 I'm glad I could help !

PR #1266, opened 2026-08-06
Music editing in the palette, MIDI import, and a recording studio

### LINKED ISSUE #1265: withRevisedNodes is quadratic in the width of a list, freezing edits to large sources
## Symptom

Splicing a node into a source that contains a wide list literal takes time quadratic in the list's length. Importing a MIDI file froze the tab with no way back; a 50KB file never resolved and had to be force-quit.

## Measurements

Appending one `Music(...)` expression to a program via `project.withRevisedNodes([[block, block.withStatement(music)]])`, where the music is a few `Track`s of `♪(…)` entries:

| notes | `withRevisedNodes` |
| --- | --- |
| 800 | 4,951ms |
| 1,600 | 19,915ms |
| 3,200 | 78,650ms |

Doubling the notes quadruples the time. At the size of `static/examples/Lyrics.wp` (8,550 notes) that extrapolates to roughly ten minutes.

Everything else in the same pipeline is fast and roughly linear, so the splice is the entire cost:

| step | 3,200 notes |
| --- | --- |
| MIDI conversion | 4ms |
| tokenize + parse | 165ms |
| `Project.analyze()` | 159ms |
| **`withRevisedNodes`** | **78,650ms** |

## Cause

`Source.clone()` regenerates preferred spacing for the replacement:

```ts
newSource.withSpaces(getPreferredSpaces(replace.replacement, newSource.spaces))
```

`getPreferredSpaces` walks every leaf of the replacement and, for each token, resolves the field its space root sits in:

```ts
root.getParent(spaceRoot)?.getFieldOfChild(spaceRoot)
```

`Node.getFieldOfChild` scans the parent's grammar and, for an array-valued field, does a linear `includes`:

```ts
getFieldOfChild(child: Node): Field | undefined {
    return this.getGrammar().find((field) => {
        const value = (this as any)[field.name];
        return Array.isArray(value) ? value.includes(child) : value === child;
    });
}
```

So for a `ListLiteral` with N elements, each of the N elements costs an O(N) scan — O(N²) overall. Nothing here is specific to music; any wide list (a long `[…]`, a large `Group` content list, a big `MapLiteral`) has the same shape.

## Possible fixes

- Memoize child → field on `Root`'s cache alongside `parents`, which is already built by the same walk.
- Or have `getFieldOfChild` consult a per-node index built once, rather than scanning per call.

Either makes the per-token lookup O(1) and the regeneration linear.

## Workaround in place

MIDI import now appends **source text** and rebuilds the source (`Source.withCode`) instead of splicing nodes, which skips space regeneration entirely: 78,650ms → 254ms at 3,200 notes, and it preserves the converter's own line layout. That only routes around the problem for one caller; every other programmatic edit to a large source still pays it.

amyjko:
# Context

Music you can edit, rather than only write. The palette gets a staff for the
`Music` under your cursor, and two other ways to get notes into a project: a
MIDI file, and your own voice.

Along the way, editing large projects got much faster. That work was not
planned — it came out of an imported song making the editor unusable — and it
is the part most worth review, because it touches `Project`, `Source`,
`Evaluator`, and the CRDT.

**Music editor.** One track at a time on a readable staff. Click to add a note,
drag to move it, arrow keys to change pitch, shift+arrows to reorder, Enter to
insert, Delete to remove. Chords, note values, add/remove track, and playback
of either the whole music or a single track, with a draggable playhead.

**MIDI import.** A `.mid` file becomes tracks of notes in their own source,
with a report of what the notation could not carry. Tempo changes used to be
dropped with a warning; they are now folded into the note lengths, so a piece
that speeds up still does. What is lost is the beat's *meaning* — a quarter
note in a 60bpm region under a fixed 120 reads as `2beats`, and `@Beat` no
longer lands on the score's beats.

**Recording studio.** Hum or sing, and the notes are written down as they are
heard. Rebuilt against a real recording after the first version transcribed
badly: the detector reports confident pitches in breath and room noise, and
loudness — not clarity — is what separates those from singing.

**Performance.** Measured, not guessed, at each step:

| Per edit, 8,000-note supplement | before | after |
| --- | --- | --- |
| `getNewConflicts` | 689ms | 0.7ms |
| `analyze()` | 606ms | 7ms |
| `getRequiredPermissions` | 53ms | 0ms |
| evaluator rebuild | 228ms | 0.1ms |
| one palette note edit | 46ms | 33ms |

The root cause of most of it: an edit builds four `Project`s
(`withSource` → `withCaret` → `bumpStampsFrom` → `withNewTime`) and only the
first carried its caches, so the one the UI rendered had none.
`Project.revision.test.ts` now fails if a bare `new Project(` reappears in an
instance method, because nothing else would.

## Related issues

-   Related Issue #390
-   Closes #1265

## Verification

`npm run check:now`, `npm test` (4,622 passing), and `npm run locales` all
clean. 29 locales translated for the new strings.

Every performance and transcription fix was **sabotage-verified**: the change
was reverted one at a time and the matching test confirmed to fail. Two tests
did not discriminate on the first attempt and were rewritten — worth knowing,
since a passing test is not evidence on its own.

Layout and drag behavior in the step-mode timeline were measured in a real
browser (Playwright against the dev server), not assumed. An earlier diagnosis
of that layout was wrong and is not in this branch.

**Not verified, and the reviewer should look:**

-   `tests/end2end/music-editor.spec.ts` and the new case in
    `commandButton.spec.ts` have not been run here. They need a build plus the
    emulator.
-   The music staff's draggable playhead and edge auto-scroll are type-checked
    and reasoned about but were not confirmed in a browser.
-   No screen-reader pass on the music editor. The notes are real buttons that
    announce their own beat, and announcements route through the central
    `Announcer`, but VoiceOver/NVDA verification is outstanding.

## Checklist

-   [x] Run the e2e suites (`music-editor`, `commandButton`) against a build
-   [x] Screen-reader pass on the music editor and recording studio
-   [x] Decide on `getPreferredSpaces`, which is 13.8ms of the 33ms remaining
        in a palette note edit — it re-spaces every token of the edited list
        when one note changed


amyjko:
CI is green — all 10 checks pass, including the shard that was failing.

The four failures were all tests asserting behavior this branch deliberately changes, not defects:

- **The palette is gated.** Three `music-editor` tests placed the caret and expected the palette to follow; it opens on demand. They open it first now. (Music makes no stage output to double-click, so the toggle is the only way in.)
- **A selected output is dashed as well as glowing**, so it reads as the same thing the editor's caret highlight does. `palette-gated-output` still expected the dashes cleared under the glow.

Fixing the first surfaced three more of the same shape: the play buttons were located by localized prose (`hear this @Music from the cursor`) and now carry a `data-uiid`; `.playhead` no longer exists because the staff draws one line, so `.cursor.playing` is what moves; and the import test asserted the notes land in the program, when the whole point of the two-source form is that they don't — it now asserts the borrow and the `Music(` that replaced them, and that the notes are *not* there.

Verified against the emulator before pushing: 37 passing across `music-editor`, `commandButton`, `palette-selection`, `palette-gated-output`, `keyboard-navigation`, and both accessibility suites.

**Not merged.** `main` requires an approving review, and I won't bypass that with `--admin`. No conflicts (`MERGEABLE`), and `strict` is off, so being 2 commits behind `main` doesn't block anything.

PR #1317, opened 2026-08-23
A complete set of built-in unit conversions

### LINKED ISSUE #363: Add missing unit conversions
## What's the problem?

Number values have numerous pre-defined unit conversions for time, distance, and  mass (see definitions in `NumberBasis.ts`. 
However, there are number of other scientific units (e.g., current, temperature, light intensity), as well as many other commonly used units, like cooking units such as fluid volume and quantities (e.g., dozen). It is possible to define custom conversions, but in most cases, any common unit someone tries should just work and be correct.

## What's the design idea?

Create a fairly complete list of units and define them for numbers.

## Who benefits?

Anyone doing unit arithmetic on common units.

## Design specification

(This section should be included after a design proposal is ready and approved, and the `buildable` tag is added. This text can remain until then. Designers should add their proposal here, not in a comment).

amyjko:
# Context

Number had 46 conversions, covering only time, length, and mass. Anything else a creator reasonably typed — a temperature, a volume, a data size, an angle in radians — produced "no conversion in scope." There are now ~220, across seventeen measures. The issue's Draft 2 was design-approved in May 2024 and never built; this implements it plus the gaps that surfaced while building it.

**The conversions are declared as data**, in the new `src/basis/UnitConversions.ts`, rather than written out as code. That was not a style preference. Hand-written blocks hide arithmetic errors: the approved design proposal itself shipped three (a horsepower that was 445.7 W in the formula and 745.7 W in the prose, a calorie equal to one joule, an imperial gallon paired with an imperial fluid ounce in an otherwise-US set), and two of the *existing* mass factors turned out to be rounded to six digits. A table puts every factor side by side and lets one generated test round-trip all of them. Each measure declares a hub unit and every other unit's ratio to it; `Convert` already searches the conversion graph, so `1km → #mi` composes through meters without a direct edge. This removed ~800 lines from `NumberBasis.ts`.

Three things beyond the approved design, each raised and agreed before building:

- **`° ↔ rad`.** `°` is already a Wordplay unit — `Place` and `Pose` rotations, `Color` hue, `°/s` velocity — and `sin`/`cos`/`tan` are `Decimal` trig, which takes radians. So `(45°).sin()` returned the sine of 45 *radians*, and there was no way to convert, because `π` is a literal and multiplying by it keeps the `°`. Those three functions also preserved the input's unit, which a ratio cannot have; they are now unitless, exactly like `arcsin`/`arccos`/`arctan` already were.
- **`s → #yr` divided by 31,449,600**, which is 52 weeks, or 364 days. It is now the Julian year, 365.25 days.
- **`? → #`** on Boolean, ⊤ to 1 and ⊥ to 0. Deliberately no conversion back: which number counts as true is the creator's decision.

Two design calls worth a reviewer's eye:

- **Volumes name their system.** There is no bare `gal`, `floz`, `pt` or `qt` — those are exactly the units that differ between US customary and British imperial, so `usgal`/`ukgal` are the two different things they could mean. Units identical in both (`oz`, `lb`, `mi`, `ft`, `in`, `yd`) stay unqualified. `tsp`/`tbsp`/`cup` are the metric-legal 5 / 15 / 240 mL rather than US customary — defensible either way, and documented in each unit's own name.
- **The draft's quantity units (`doz`, `scr`, `M`, `B`) are dropped**, per the unanswered question on the issue: unitless numbers already count anything, and `B` is better spent on *byte*.

## Localization

Each unit's name is translated once (`basis.Number.unit.*`, 125 short words) and filled into a single template, rather than translating 220 sentences — that turned ~4,000 `$?` placeholders across the locales into 127. The 46 per-conversion doc keys are retired.

The template is **`$from → $to`**, not a sentence. A bare preposition between two nominative nouns is ungrammatical in a large share of these languages: Kannada and Assamese were rendering case suffixes as free-standing words, and Serbian, Greek, Marathi, Nepali, Tagalog, Turkish and Tamil all read as a *range* or a *ratio* rather than a conversion. The arrow needs no grammar and is Wordplay's own conversion operator, so the doc for `#m → #km` reads "meters → kilometers". Unit names are kept rather than falling back to symbols, since `km` doesn't tell a learner what a kilometre is.

**The machine translations were audited per-locale, and 25 names were wrong — see the [audit comment](https://github.com/wordplaydev/wordplay/pull/1317#issuecomment-5388286311)** for what was fixed and, importantly, the named list of items that still need a native speaker.

## Related issues

- Closes #363

## Verification

- `npm run check:now` — 0 errors, 0 warnings across 3,144 files.
- `npm test` — 396 files, 7,251 passing.
- `npm run build` — clean.
- `npm run locales` — exits 0; all locales regenerated with `npm run locales-fix`, schemas with `npm run create-schemas`, prettier over every changed JSON.

New `src/basis/UnitConversions.test.ts`, 156 tests:

- **A generated round trip for every hub↔unit pair**, built from the table so a newly added unit cannot skip its own test.
- **~40 known values written out by hand** against published definitions (SI brochure 9th ed., NIST SP 811 App. B, the Weights and Measures Act), *not* derived from the table — a round trip only proves the two directions agree, so a mistyped factor round-trips happily. Each was recomputed in arbitrary precision independently of the code.
- **Composed paths with no direct edge**: `1kWh → #cal`, `1K → #°F`, `1usqt → #ukfloz`, `1ly → #au`, `1KiB → #b`.
- The trig case: `(45° → #rad).sin()` is correct and carries no unit.

Two existing assertions in `NumberValue.test.ts` changed, because the old factors were wrong: `1kg→#oz` was `35.274oz` (six-digit rounding) and is now `35.273961949580412916oz`; `1000mg→#lb` likewise.

**An earlier revision of this description blamed an intermittent test failure on `drift.test.ts`. That was wrong.** Captured properly, it was `Basis.test.ts` *timing out* rather than failing: it deliberately defeats the basis cache and builds two cold bases at ~1.5s each, against vitest's 5s default. Cold basis construction measures 1504ms on this branch and 1465ms on main, so the ~220 conversions are not the cost — the allowance was always too tight for what that test does, and it now has an explicit one.

### Accessibility

No UI was added or changed — this is entirely language-level, so there is no new surface to keyboard-test, no new color, and nothing new for a screen reader to read. The generated conversion docs render through the existing guide markup path, verified in the locales whose templates were previously broken.

## Checklist

Two consequences measured but not fixed, both better as your call than as scope I add myself:

- [ ] **The units menu grew from 31 entries to 126.** They land in one collapsible `RevisionSet` rather than 126 top-level rows, and typing `1km` directly is unaffected — the menu is a discovery aid. But it is noticeably longer, and the menu has no text filter. Follow-up issue?
- [ ] **Analysis is slower per `→`.** `Type.getAllConversions` returns *every* conversion in the basis, not the ones on that type, so the BFS's per-node `filter` scales with the whole list — which went from 65 to 237. Measured: a program with no conversions 0.47ms → 0.55ms; a deliberately conversion-dense one (6 conversions in 8 lines, one unresolvable so the BFS must exhaust the graph) 2.0ms → 3.6ms. The fix, if it ever matters, is to bucket conversions by input basis-type name and — for numbers — by input `Unit`, which is pooled and canonical, making the lookup O(1). Left alone rather than adding that machinery to save under 2ms.
- [ ] **Native-speaker review** of the items in the audit comment — `atm` in the Indic locales, the Serbian genitive-plural block, and the Peninsular-vs-Mexican Spanish register call are the three worth a person.

`LANGUAGE.md` gains the full unit table and the three rules that need saying out loud (temperature shifts as well as scales, volumes name their system, `°` and `rad` measure the same thing). `ARCHITECTURE.md` notes that the conversions are data. `CHANGELOG.md` has four Added and three Fixed bullets.


amyjko:
## Translation audit

The 127 new strings were machine-translated via `translate.yml`, then audited across all 29 locales — mechanically for defects a script can decide, and by a per-language reviewer for meaning. The mechanical pass found 2 problems in 3,625 names; the language pass found considerably more, so the summary below is what a reviewer should actually trust.

Before translating, `degrees` was renamed **`degrees of angle`**. The translator receives bare strings with no key path or field context and chunks them 25 at a time in path order, so `degrees` was arriving between `kelvins` and `radians`, where it reads as a temperature. Everything else got usable context from its chunk — `stones` sat between `pounds` and `US short tons`, `bars` between `kilopascals` and `atmospheres`.

### The doc template was the systemic problem

`"$from to $to"` assumes a language where a free preposition sits between two nominative nouns. That is false for a large share of these locales, and reviewers flagged it independently in fifteen of them:

- **Case suffixes rendered as free-standing words.** Kannada produced `ಮೀಟರ್‌ಗಳು ಇಂದ ... ಗೆ` — `ಇಂದ` and `ಗೆ` must agglutinate (`ಮೀಟರ್‌ಗಳಿಂದ`, `ಮೈಲಿಗಳಿಗೆ`) and read as broken text spaced out. Assamese `ৰ`/`লৈ` the same. Telugu `నుండి` needs an oblique stem the nominative values don't have.
- **Reads as a range or a ratio, not a conversion.** Serbian `до` ("up to"), Greek `προς` (a ratio, "3 προς 5"), Marathi `ते`, Nepali `देखि` (dangles without `सम्म`), Tagalog `hanggang` ("until"), Tamil `முதல்…வரை`.
- **Wrong word.** Turkish `ile` means "and" — `kilometre ile mil` reads as a pair, not a conversion.
- **Orthography.** Hebrew hyphenated `ל-` before a Hebrew word, where the prefix must attach directly; Korean spaced particles that must bind.

It is now **`$from → $to`** in every locale. The arrow needs no grammar, and it is Wordplay's own conversion operator, so the doc for `#m → #km` reads "meters → kilometers". Unit *names* are kept rather than falling back to symbols — `km` doesn't tell a learner what a kilometre is, and symbols would duplicate the signature the guide already shows.

### 25 names were wrong, and are fixed

Each was confirmed against the file, not taken on the reviewer's word:

| Kind | What was wrong |
| --- | --- |
| Wrong quantity | zh-CN kibibyte = `千二百八十字节` ("1280 bytes"); el-GR milliwatt = `χιλιοβάτ`, which is what its own kilowatt says (off by 10⁶); vi-VN deciliter = "cubic decimeter"; id-ID ounce = `ons`, which is 100 g in Indonesian; he-IL watt hour = "watt per hour" — the same construction that locale uses for kilometers per hour |
| Wrong unit entirely | `acre` became a different local land measure in six locales: Morgen (~2500 m²), στρέμματα (1000), dönüm (1000), tunnland (4937), فدان (4200), and hektar — which already named the hectare |
| Collisions | tr-TR gave knots and nautical miles the same name (a knot is one nautical mile *per hour*); Tamil kibi and gibi transliterate identically, since `க` serves both /k/ and /g/, so those two now carry their symbol |
| Defects | ro-RO `kiloocteți` had lost its `o`; sr-RS `ампери` ended in a Latin `i` and `centilitri` was Latin script in a Cyrillic file; he-IL bytes = `בתים` ("houses"); `stones` came out as rocks in Spanish, Polish and Swedish |
| Wrong sense | ko-KR, zh-CN and zh-TW measured bpm in *heartbeats* — Wordplay's bpm is musical tempo |

Note that "stones as rocks" is exactly the English-ambiguity failure this design was most exposed to, and chunk context caught it in most languages but not all three.

### Left for native speakers

These are grammar, register and consistency judgements I'm not in a position to make, and I'd rather flag them than guess:

- **`atm` reads as the everyday word for "the weather" or "the sky"** in bn-BD, as-IN, gu-IN, pa-IN, te-IN, kn-IN and mr-IN — "pascals → the environment". The clearest item on this list, and the same mistake in seven locales.
- **sr-RS has a 25-key block in the genitive plural** (`вата`, `џула`, `паскала` …) where the rest of the file is nominative, plus `bar` = `бара`, which is also the everyday word for a puddle. Also `uston`/`ukton` dropped *short*/*long*.
- **es-MX uses Peninsular forms** (`julios`, `vatios`, `hercios`, `amperios`) where Mexican usage per NOM-008-SCFI is the international ones.
- **Rate units are calqued into English word order** in ta/te/kn (`மீட்டர் ஒரு வினாடிக்கு`), each contradicting that same file's own `bpm`.
- Smaller: tl-PH `antas` ("rank") for degree and English plurals on 14 borrowings; ne-NP `हर्ज` (a word meaning "harm") for hertz and `BTU` spelled `ब्रिटिस`; ja-JP invented `アメリカトン`/`イギリストン`; el-GR five untranslated Latin strings; several transliteration splits within one prefix family.

Many unit names were deliberately **left in English** and that is correct, not a gap — `radians`, `hectares`, `joules`, `watts`, `volts`, `hertz`, `lux`, and the whole KiB/MiB/GiB/TiB family are genuinely the same word in most of these languages.

### Verification

`npm run locales` exits 0, `npm run check:now` is clean, `npm test` passes 7,152, and the rendered docs were checked in the previously-broken locales:

```
en-US: nanoseconds → seconds | metric tons → grams | degrees Celsius → degrees Fahrenheit
kn-IN: ನ್ಯಾನೋಸೆಕೆಂಡ್‌ಗಳು → ಸೆಕೆಂಡುಗಳು | ಮೆಟ್ರಿಕ್ ಟನ್‌ಗಳು → ಗ್ರಾಂಗಳು
sr-RS: наносекунде → секунде | тоне → грами
he-IL: ננושניות → שניות | טונות מטריות → גרם
tr-TR: nanosaniye → saniye | metrik ton → gram
```

One caveat worth stating: three of my local full-suite runs reported a single failure that never reproduced and whose name I failed to capture before it scrolled; four subsequent runs, including two under the JSON reporter, were clean, and `unit / vitest` was green on the previous commit. If CI reproduces it, it is worth chasing rather than dismissing.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #140, opened 2023-08-21
Hindi locale


amyjko:
**What's the problem?**
We need a Hindi locale.

**Who benefits?**
All fluent Hindi readers; ~7% of the world. Use machine translation to start, then edit.

yshastri5:
@amyjko I would like to work on issue. it would be amazing if you can add me to the issue.

adhimanjagota:
Hi Amy, 
I'm quite fluent in Hindi and would like to resolve this issue. Can you please assign it to me?

adhimanjagota:
Hey! I've created the folder to resolve the issue. Here's the location - wordplay/static/locales/main/hn-IN

amyjko:
Added you both!

adhimanjagota:
Hey! We have completed the localisation. Should I just upload it to the folder?

amyjko:
@adhimanjagota All contributions should be submitted as pull requests, per the workflow defined in the wiki. Read the localization wiki for instructions.

amyjko:
Partially resolved by PR #622.

tris-23:
hi @amyjko can I get added to this issue? I would like to recheck some of the translations.

ananyat24:
Hi @amyjko! Can I get added to this issue as well? Thank you!

amyjko:
Added you both.

@adhimanjagota @yshastri5, are you still contributing to this? Please let us know one way or another, so we don't have people doing duplicate work.

adhimanjagota:
@amyjko yes we're still contributing to this issue and will love to add a value to the community. We'll be attending the meet up next week to divide up the work with @ananyat24

amyjko:
Thanks @adhimanjagota! Coordination is key; I recommend working in a shared branch on this, so you don't do redundant work and can identify conflicts early.

daminihs:
Hi @adhimanjagota and team working on this, I reviewed the Hindi JSON file and noticed a couple of discrepancies. I’m not sure if work has already started on these or if there’s a specific reason they’re structured this way, but I thought I thought I’d check with you all.
1.  "Stream" translation in 'term' is different than 'token '- token has the correct translation (dhara).
2. "Language" is  translated to "bhasha tag" - not sure if the ~tag~ was intentional
3. "TypeOperator" is just "hai" - incorrect
4. "table" is translated to a dining table "mez". should be something like तालिका . and should be consistent with "TableOpen" / "TableClose"
[...]

PR #1319, opened 2026-08-23
140 review hindi locale – pull #2

ellipsis7449:
# Context

Changes to Hindi locale, including glossary terms and definitions; documentation; and other areas of the locale JSON.

## Related issues

- Related Issue #140

## Verification

I've verified many of my changes by running `npm run dev` and finding the areas where my changes should occur.

## Checklist

- [ ] Have another Hindi speaker review my translations.


amyjko:
Thank you for this, @ellipsis7449 — **your review is merged.** It's on `main` in 70153752d.

Closing this pull request, but not because the work wasn't wanted: we reorganized how locale files are stored, and that made this branch impossible to merge in the ordinary way.

**What changed.** A locale used to be one file of about a megabyte, `static/locales/hi-IN/hi-IN.json` — the file this branch edits. It's now eight smaller files in a `sections` folder. That was done precisely because of branches like this one: when everyone edits the same enormous file, a branch that stays open for a while becomes nearly impossible to merge, and yours had been open a long time.

Which left this branch pointing at a file that no longer exists. Git would have reported a conflict on every line you changed, including the ones nobody else had touched.

**So we moved it for you rather than asking you to redo it.** Your changes were reapplied to the new layout by locale path rather than by line, with you kept as the author. **232 of your changes landed.**

**Two didn't, and you should know which:**

- `ui.docs.mode.purpose.labels` and `ui.page.galleries.section.tabs.tips`. These are positional lists — each slot means a particular thing, so the list has to be exactly as long as the English one. English gained a slot in one and lost one in the other while this branch was open, so your versions no longer fit. Nothing is wrong with your text; it was written for a shape that changed underneath it.

We also ran the repair tool afterwards, which restored some glossary links (`@value` and similar) that the older text had lost.

**Two things left that need a person who reads Hindi**, and we don't auto-fix either, because both are judgment calls about words:

- `Expression.place` is `जगह` while the type `Place` is `स्थल`. These should normally be the same word, the way English uses "place" for both.
- Six glossary words no longer appear anywhere in the locale's own writing: `wordplay` (शब्दकला), `expression` (समीकरण), `scope` (कार्यक्षेत्र), `how` (निर्देशक पुस्तिका), `placeholder` (खाली चीज़), `index` (सूचकांक). A `@reference` to one of those can't explain itself if the word is never used. Either the glossary word or the prose should change — and which one is a translator's call, not a script's.

If you'd like to keep going, both of those are good next steps, and you can now edit Hindi directly from the site at https://wordplay.dev/localize, which opens a pull request for you — no git needed. The files are also much easier to work with now.

Thank you again for the review, and sorry it took a reorganization to get it in.

ellipsis7449:
Sounds good, thank you Amy! I'm sure the smaller files will be much easier to work with, and I'll work on those two issues you mentioned in the fall.

Issue #220, opened 2023-10-24
Vertical right to left layout doesn't make sense for horizontal left-to-right languages


willGG11:
## What's the problem?

Need to consider a more practical scenario when designing the writing layout feature. 
In our beta version, go to the settings section
and change the writing layout from horizontal to vertical, only making the content flip 90 degrees. Makes users hard to access text. (They had to crane their necks to read the content) and less sense-making.

## What's the design idea?

I think we could optimize the writing layout into reading reading-friendly version.
-----------------------
an example case:
this is a demo (horizontal view)

changing into our new vertical view:

this 
is 
a 
demo
-----------------------
One scenario based on my cultural background:
Chinese have both horizontal and vertical reading and writing habits in different regions
In mainland China, we typically use a horizontal layout:
我喜欢读书
In some regions of China, people prefer a vertical layout, especially in published books:
我
喜
歡
讀
書

It is important to get both layouts accommodated and optimized in the Chinese version of wordplay.
current implementation can't successfully achieve this feature. 
<img width="674" alt="截屏2023-10-23 22 03 52" src="https://github.com/wordplaydev/wordplay/assets/106497657/beb66437-ab46-4e81-8a79-165206206835">
<img width="999" alt="截屏2023-10-23 22 03 24" src="https://github.com/wordplaydev/wordplay/assets/106497657/deae02da-01e6-45c2-9386-f5e08eb2df52">

## Who benefits?

Is this something you think only you would benefit from? Are there others that would benefit? Are there others that might be harmed by supporting this?

Different languages have different reading and writing habits, and this change will benefit people who write and read vertically all over the world. At the same time, it is important to ensure that this change does not affect people who write and read horizontally.

amyjko:
Thanks @willGG11. I'm having a bit of trouble making sense of the design proposal here. Are you proposing that left-to-right, top to bottom languages like English have a vertical reading mode? Or are you proposing to properly support vertical layout for CJK (Chinese/Japanese/Korean) languages?

If the former, yes, we currently provide some unhelpful configurations for Latin scripts, and it might make sense to disallow the ones that don't make sense. If you meant the latter, than this issue should focus on any defects you've found in CJK support.

willGG11:
[...]

PR #1330, opened 2026-09-03
Full writing-mode support for the editor and UI prose (#1203)

### LINKED ISSUE #1203: Full vertical writing-mode support for the editor and UI prose
## Summary

We now have a `writingLayout` setting (`auto` | `horizontal-tb` | `vertical-rl` | `vertical-lr`) that **program output** respects, and genuinely-vertical scripts (Mongolian `Mong`, Phags-pa `Phag`) are classified `vertical-lr`. But the **code editor** and **UI prose** remain horizontal-only. So a creator who selects a vertical layout (or, in the future, a vertical-dominant locale) gets vertical *output* inside a horizontal *editor and interface* — inconsistent, especially for uniformly-vertical scripts where a mixed layout reads as broken rather than bilingual. This issue captures that debt.

### 1. Vertical editing in the code editor
`CaretView.svelte` hard-codes `const horizontal = true;`, and caret position, selection rectangles, and arrow-key navigation all read physical `.left/.top/.right/.bottom` from `getBoundingClientRect()`. Applying `writing-mode: vertical-*` via CSS renders text vertically but leaves editing broken (caret lands in the wrong place; arrows move the wrong way). Functional correctness needs a geometry rework that maps physical ↔ logical (inline/block) axes throughout:

- `src/components/editor/caret/CaretView.svelte` — caret/selection geometry (the bulk of the work)
- `src/components/editor/util/rowModel.ts` — row/column clustering (axis inversion)
- `src/components/editor/commands/Commands.ts` — arrow-key → direction mapping (add the vertical axis alongside LTR/RTL)
- `src/components/editor/pointer/PointerUtilities.ts` — pointer hit-testing / insertion points
- `src/components/editor/Space.svelte` — soft-wrap measurement under vertical flow

### 2. A coherent "vertical reading mode" for UI prose
`MarkupHTMLView.svelte` now carries the locale's `lang`/`dir`, but prose layout stays horizontal. Just flipping `writing-mode` on the markup is not enough: prose sits inside hosts built horizontally — the `Speech.svelte` bubble (absolutely-positioned border-geometry tail), `Dialog.svelte` (`width: 95vw`, column flex), chat rows — so a vertical block grows/scrolls along the wrong axis, the speech tail renders incorrectly, and `scaleY` paragraph animations end up perpendicular to text flow. A real vertical reading mode would reorient the reading surfaces (Guide/Docs/Tutorial panes), adapt or replace the Speech tail, and flip scroll axes.

## Out of scope (by design)
Purely spatial chrome — toolbars, the tile window manager, sliders, color/font pickers — should **stay horizontal** regardless of writing layout; vertical chrome isn't sensible. The directional respect those need (RTL mirroring) is already implemented.

## Notes
- Already in place to build on: the `writingLayout` setting, `RenderContext.layout`, and per-`Phrase` `direction` (which defaults to inherit) — see `src/db/settings/WritingLayoutSetting.ts`, `src/output/RenderContext.ts`, `src/output/Phrase.ts`.
- This is a design-led, multi-week effort; the editor caret geometry is the hard part.
- Web convention for uniformly-vertical scripts (e.g. Mongolian) leans toward full-page vertical (reorienting reading surfaces), not vertical text stranded in horizontal chrome.
- Related: #220.

amyjko:
Closes #1203. Someone who reads a language written down the page now gets an interface and an editor that do the same — and the two are decided **separately**, which was the design mistake this started from: the interface is in one or more globally declared languages, but code can be in any combination of them, including different ones.

## What changed

**The interface follows the reader.** A `writingLayout` setting, offered only when a chosen locale's dominant script has a vertical tradition, and then only in the direction that script is actually set in — CJK runs its columns right to left, Mongolian and Phags-pa left to right, and no script uses both. `resolveWritingLayout` is the only thing anything reads, so dropping the locale that offered the control also stops it applying without forgetting the choice.

**The editor follows the code.** `eligibleWritingLayouts` reads a source's own *glyphs* and returns what it could be set in; the toolbar offers those and the creator picks, per project and source. A Latin program stays horizontal however its reader reads. Four traps make the naive version silently useless and each has a regression test: `ƒ`/`ø` are Latin letters, a unit's `m` lexes as a `Name`, `ー` is `Script=Common` (so `\p{scx=}`, never `\p{sc=}`), and `Hans`/`Hant`/`Kore` aren't Unicode script values at all.

**The editor's geometry is inline-axis geometry.** `axes.ts` projects viewport rectangles onto a logical basis — inline along the text, block across the lines, both increasing in reading order in every mode — so caret placement, row clustering, selection outlines and hit-testing keep every `<`, `Math.min` and sort comparator they had. RTL stops being a separate mechanism: a right-to-left inline axis is a negated projection. Arrow keys are *remapped*, not re-bound, so every command keeps its behavior and its localized description in 30 locales.

**Output is deliberately untouched.** It follows the *project's* locale, so a composition looks the same to everyone, and a creator makes a phrase vertical with its own `direction` — the design #220 settled on.

Also fixes the front page's Login link, which sat at the page edge instead of lining up with the route grid below it (pre-existing, unrelated to the rest).

## The audit

The last pass over this was an audit of every CSS, scrolling and layout change, paying particular attention to routes not examined while building it. It found two real regressions, both of which every existing test passed straight through:

- **`RootView` declares `horizontal-tb`** so a code snippet inside documentation is never laid sideways by the reader's prose setting — but the editor renders its code through `RootView` too, so that rule reached inside and forced the code back horizontal **while the editor element itself still reported vertical**. The vertical editor was entirely dead and every test passed, each of them measuring the container rather than the code. The editor now passes a `layout` prop.
- **The reading surface was on `Writing.svelte`**, which backs eighteen routes — and most of them are listings (`projects`, `galleries`, `characters`, `teach`) or forms (`login`, `join`, `profile`, `localize`), places you *operate*. Marking the wrapper turned all of them vertical and collapsed `/projects`' grid to **zero pixels wide**, which is the same column rule that had already bitten the concept listing. It is now opt-in and six prose pages take it.

And three costs the projection introduced but doesn't have to pay: `Axes` grew scalar `blockStart`/`blockEnd`/`containsBlock` because the pointer code tests every token view in the source on every move of a drag and projecting before filtering turned a zero-allocation comparison into three objects per token; `eligibleWritingLayouts` is memoized on the immutable `Source`, which the editor and its toolbar both derive on every keystroke; and the paragraph reveal went back to two keyframe sets rather than one driven by custom properties inside `transform`. One gratuitous change was reverted — chat's scroller is never a reading surface, so `overflow-block` bought nothing and cost engine support.

The tutorial's vertical layout was also rebuilt, because the first attempt had it backwards: **vertical text takes its line length from the block's *height***, so the full-height column the dialog already occupied is the right container (a ~47-character line, where stacking it into a wide strip gave 12 and squeezed the lesson to 143px).

## Verification

- `npm run check:now` — 3377 files, 0 errors, 0 warnings
- `npm run rtl` — clean, and now enforced in `npm test` via `logicalCssSync.test.ts`
- `npx vitest run` — 10,869 passing
- e2e (chromium, against a fresh build): `writing-vertical`, `caret-vertical`, `accessibility`, `accessibility-authed`, `keyboard-navigation`, `aria-snapshots`, `landing-nav-handover`, `chat`, `project`, `fold`, `search`, `editor-output-flow`, `tutorial-tour`, `stage-selection` — 138 passing

Real screen-reader verification (VoiceOver/NVDA) has not been done and is the main thing left for review.


amyjko:
Added `13b73f7` (cherry-picked to both branches): a locale named in the URL lost a race with the settings sync.

`syncUser` reads the creator document and applies every setting in it, locales included, and it lands whenever the network answers — after hydration has already applied the locale the URL asked for. A signed-in creator following a link in another language saw that language render and then flip back to their own, at a moment decided by how fast Firestore replied. That is what failed here: `validation-rtl`'s two Hebrew tests assert on the *served* markup first (which passes) and then on the field's computed direction (which by then is `ltr`).

**It is not new, and it is not this branch's doing** — the same file list fails identically on `main`. What each of these branches does is add a spec file, which reshuffles the shards and lands `validation-rtl` under load, where the race is lost. Which is why the fix is on both: whichever merges first, the other would have hit it.

`LocalesDatabase` now latches whether the locale came from the URL, synchronously before anything is awaited, and `syncUser` leaves the locales alone when it did. `localePath`/`localeGoto` build every internal link from the URL's own locale, so the latch cannot pin a locale the reader did not arrive with. One deliberate behavior change: a signed-in creator opening a bookmarked `/en-US/…` link now stays in English for that page load rather than switching to their account's language after a beat.

No new test: the failure is a race a test can only lose on purpose, and a version that passed against the known-broken build would be worse than none — I wrote one, confirmed it passed *without* the fix, and deleted it. The covering test is `validation-rtl`, which goes from failing to passing in the full shard (7 failures → 2, and both remaining fail identically on `main` from local emulator state).

PR #1331, opened 2026-09-03
A concept goes stale when the reader changes locale

amyjko:
Reported: switching locales in the guide makes the links on the right of a concept view disappear; refreshing restores them.

## What was happening

The guide's navigation history holds the concept you are reading as an **object**, and changing locale rebuilds the concept index with entirely fresh ones. Nothing remapped them, so the concept went on rendering — it carries its own documentation — while everything the *new* index was asked about it came back empty: `getHowTosForConcept` is a `Map` keyed by concept identity, so the how-to links simply vanished.

It is a little worse than reported, which the regression test caught: the prose stayed in the language you had just **left**, too. The guide stays mounted across the switch (SvelteKit reuses the component for the same route, and `afterNavigate` only re-resolves the history on `popstate`), so nothing re-read the history from the URL.

`ProjectView` already remapped its path on rebuild — but with `getEquivalent`, which asks `isEqualTo`. A basis is built per locale, so the definitions a concept wraps are different objects: it recognized **121 of 902** concepts across a locale change and silently *dropped* the rest, sending the docs tile home. Same bug, different mask.

## The fix

`ConceptIndex.getCorresponding` tries structural equality first — exact when the index was rebuilt for the same locales, which is what a project does on every edit — then falls back to the concept's URL token, which resolved **902 of 902** in both directions. A token is a concept's character before its name, and a character is the same word in every language.

Two things are load-bearing, one of which cost me an iteration:

- **The remap compares by object identity, not `sameHistory`.** `sameHistory` asks the very `isEqualTo` that calls a stale concept and its replacement the same thing — so the first attempt computed the remap and then threw it away, leaving the bug exactly as it was while passing every unit test I had written for it. That is why the regression test drives the interface rather than the index.
- **Concepts are indexed by token.** A project rebuilds its index on every edit and remaps the whole path each time, and a linear scan per concept made that quadratic. A miss still falls through to the exhaustive match, so a URL written with one of a concept's other names still resolves.

## Verification

- `npm run check:now` — 3379 files, 0 errors, 0 warnings
- `npx vitest run` — 10,873 passing, including `conceptRemap.test.ts` (every concept's token resolves in both directions, and structural equality alone demonstrably does not)
- e2e: `guide-locale`, `project`, `accessibility`, `accessibility-authed`, `search`, `writing-vertical` — 81 passing, and `guide-locale` run three times over for flakiness
- **Confirmed the new e2e test fails without the fix** rather than passing vacuously

Unrelated to and independent of #1330, so this branches from `main` — but note both touch `ProjectView.svelte`, so whichever lands second will want a look.


amyjko:
Added `13b73f7` (cherry-picked to both branches): a locale named in the URL lost a race with the settings sync.

`syncUser` reads the creator document and applies every setting in it, locales included, and it lands whenever the network answers — after hydration has already applied the locale the URL asked for. A signed-in creator following a link in another language saw that language render and then flip back to their own, at a moment decided by how fast Firestore replied. That is what failed here: `validation-rtl`'s two Hebrew tests assert on the *served* markup first (which passes) and then on the field's computed direction (which by then is `ltr`).

**It is not new, and it is not this branch's doing** — the same file list fails identically on `main`. What each of these branches does is add a spec file, which reshuffles the shards and lands `validation-rtl` under load, where the race is lost. Which is why the fix is on both: whichever merges first, the other would have hit it.

`LocalesDatabase` now latches whether the locale came from the URL, synchronously before anything is awaited, and `syncUser` leaves the locales alone when it did. `localePath`/`localeGoto` build every internal link from the URL's own locale, so the latch cannot pin a locale the reader did not arrive with. One deliberate behavior change: a signed-in creator opening a bookmarked `/en-US/…` link now stays in English for that page load rather than switching to their account's language after a beat.

No new test: the failure is a race a test can only lose on purpose, and a version that passed against the known-broken build would be worse than none — I wrote one, confirmed it passed *without* the fix, and deleted it. The covering test is `validation-rtl`, which goes from failing to passing in the full shard (7 failures → 2, and both remaining fail identically on `main` from local emulator state).

PR #1348, opened 2026-09-06
Make the test suites fast, and fix the query bug behind the e2e flakes

amyjko:
# Context

The test suites had become the bottleneck in local iteration and in CI. Chasing that turned up a real app bug, so this PR is both.

Both suites roughly doubled in the last month — e2e specs 33 → 66, unit test files 269 → 513, about ten new files a day — so alongside the mechanical work there is a written policy for where a test goes and when to run the suites.

## The app bug

Firestore allows exactly one `array-contains` clause per query, **including inside an `or()`**. Three queries carried two or more, so Firestore rejected each outright and the listener never delivered a snapshot:

| Query | What it was for |
| --- | --- |
| `GalleryDatabase.startSync` | the signed-in creator's galleries listener |
| `ProjectsDatabase.syncUser` | the base owned/shared projects listener |
| `CharacterDatabase.getByName` | the `@oldname/Character` alias fallback |

Nothing surfaced because galleries, projects and characters still appear — from the Dexie cache and from explicit reads. What was lost was every *live* update: a gallery just created never reaching the share picker, a how-to space rendering empty, membership changes never arriving. That reads as flaky sync rather than as a broken query, which is why it lasted. It affects production, not only tests.

Confirmed from the browser console, which logged eight `Only a single array-contains clause is allowed in a query` errors per page load and now logs none. Each query is now one listener per clause, with removal decided against the union of all of them — the rule `CharactersDatabase` already applies.

## Measured effect

| | Before | After |
| --- | --- | --- |
| `npm run test:run` (local) | 77s | **15s** |
| Full chromium e2e (local, 4 workers) | ~7 min | **2.9 min** |
| e2e tests | 303 | 267 (same coverage) |
| Flaky failures per e2e run | 2–3 | 1 |

The unit win is one structural change: five files whose cost scales with the committed corpus rather than with the code were 172 of the suite's 318 CPU-seconds, and one of them *was* the wall clock. They move to a `sweep` project with its own CI job and a pre-commit gate. The e2e win is fewer navigations — the axe specs now visit each surface once and scan both colour schemes on it.

## Related issues

- Related Issue #
- Closes #

## Verification

- `npm run check:now` — clean, 0 errors 0 warnings.
- `npm run test:run` — 8668 passed. `npm run test:sweep` — 2488 passed. Both also pass under `--sequence.shuffle.files` with a cold `.vite/vitest` cache, which is the check for the ordering hazards a project split can introduce.
- `npm run locales` — exit 0.
- `npx prettier --check .` — clean.
- Full chromium e2e: 266 passed, 1 failed (see below).
- The axe merge was verified before being relied on, by comparing **full axe result objects** between a fresh navigation in a scheme and a page flipped into it, across all 13 public routes and the project editor with a running stage. They match; the one route that differed differed only in axe's non-gating `incomplete` list.
- The `array-contains` fix was verified by capturing browser console errors in a narrowed repro before and after.
- Worker count is a measurement, not a guess: over the eleven cloud-assertion tests that contend hardest, 2 workers failed 2, 4 failed 1, 7 failed 4.

Not verified: no screen-reader or keyboard pass, because there is no user-visible UI change here — the only app changes are Firestore query construction. The axe gate itself still runs over every surface in both schemes.

## Checklist

- [ ] **One flake remains and is not root-caused.** `howto-form.spec.ts:70` still fails intermittently. It passes 3/3 alone and fails when other gallery specs have run, and the console is now clean, so it needs a fresh dig rather than a longer timeout. Previously this spec failed with a 90s timeout; it now fails in ~20s.
- [ ] `expect: { timeout: 10_000 }` is a global Playwright change I have not exercised against every spec.
- [ ] The write-serialization half of the first commit fixes **no observed failure** — it is a latent lost-update path found while chasing the flakes, with a test that fails without it. Reasonable to drop if you would rather keep the diff tight.
- [ ] CI timings on this PR are the thing to watch: the `vitest` job should roughly halve, a new `sweep` job runs beside it, and the playwright shards should come down.


amyjko:
## Measured CI impact

Comparing this PR's run ([34063319197](https://github.com/wordplaydev/wordplay/actions/runs/34063319197)) against a recent baseline PR run ([34014201761](https://github.com/wordplaydev/wordplay/actions/runs/34014201761)):

| Job | Before | After |
| --- | --- | --- |
| `unit / vitest` | 5m36s | **2m31s** |
| `unit / sweep` | — | 3m38s (new, parallel) |
| `unit / locales` | 3m41s | 4m05s |
| `playwright 1/3` | 5m53s | 6m32s |
| `playwright 2/3` | 6m27s | 5m52s |
| `playwright 3/3` | 6m55s | 6m16s |
| **Whole gate** | **7m00s** | **6m32s** |

### The honest read

The vitest job did what it was supposed to — less than half — but **the gate is playwright-bound and this PR barely changed that**, about 28 seconds. Worth being explicit that the local developer loop is where the win landed: `npm run test:run` 77s → 15s, and a full local chromium e2e ~7 min → 2.9 min.

Three reasons the e2e shards didn't move much, none of which this PR addresses:

1. **CI still runs `workers: 2`.** The local default went to 4 on measurement; CI's value comes from `playwright.yml` and was deliberately left alone.
2. **~80s per shard is fixed preamble** — checkout, `npm ci`, browser cache, JDK, functions install, and a 40s build that is identical across all three shards.
3. **Shard balance got worse, not better.** Playwright shards by file, and `accessibility.spec.ts` went from 34 tests to 17 — so the shard holding it got lighter while the others didn't, and the gate is the *slowest* shard. Shard 1 actually got slower.

Shard 3 also spent ~40s on retries for the two flaky tests below.

### Next lever, if the CI gate matters more than the local loop

Raising CI workers, or sharding by measured duration rather than by file. `reportSlowTests` (added here) now names anything over 20s, which is the input that was missing.

### Flake status

`howto-form.spec.ts:70` failed and was retried; `palette-gated-output.spec.ts:126` was flaky and passed on retry.

I dug into the first one and want to record what it is *not*, since three plausible theories are now ruled out by measurement: it is not concurrency (fails at 1, 2 and 4 workers alike), not the Firestore transport (forcing long polling changed nothing), and not gallery accumulation (25 galleries, no effect). Fixing the `array-contains` queries removed most of its failures but not all. On CI it now fails on the `user-select` poll rather than on the tile wait, and the captured screenshot shows an empty canvas and an "unsaved" chip.

That last detail points at a real deadlock I found but am **not** fixing here, because a change I made for it did not reduce the failure rate and I could not write a test that fails without it: `GalleryDatabase.edit` marks a gallery dirty in Dexie *before* the cloud write but mirrors the gallery itself *after* it, so an interruption in between leaves a dirty flag for a gallery Dexie has no copy of — after which `seedDirty` re-adds the id to `unsavedIDs`, the listener's skip-dirty guard refuses the server copy for the rest of the session, and `flushUnsaved` has nothing local to re-push. It deserves its own change with its own test rather than riding along here.

amyjko:
## Root cause found and fixed — CI fully green

[Run 34065915260](https://github.com/wordplaydev/wordplay/actions/runs/34065915260): all nine jobs green, **all three playwright shards passing with no retries**.

### What it was

`firestore.rules` allows a gallery update only if `moderation`, `moderatedAt`, `flags` and `words` are absent from the write or identical to what is stored. `words` is rebuilt by the `galleryEdited` trigger on *every* change — so a client that writes the whole document ships whatever `words` it last saw, and the moment the trigger has moved on, Firestore denies the write with `permission-denied`.

That denial is silent, and it wedges the gallery:

1. `trackSave` records the failure and leaves the gallery in `unsavedIDs`.
2. `seedDirty` restores that flag on every reload.
3. The listener's skip-dirty guard then refuses **every** server snapshot for that gallery for the rest of the session.
4. `flushUnsaved` retries the same denied write forever.

Observed directly, by logging the listener's skip branch and `trackSave`'s catch:

```
[DX] trackSave FAILED galleries 392fb45a… FirebaseError: [code=permission-denied]
[DX] listener SKIPPED dirty gallery 392fb45a… howTos= ["4ffa1ecc-…"]
```

The client was receiving the gallery **with the how-to in it** and throwing it away. Hence an empty how-to canvas, a share picker missing an option, and the "unsaved" chip visible in the CI screenshot.

This is not a test problem. For a creator it means renaming a gallery can stop that gallery updating on their device until they clear site data, and `flushUnsaved` will keep pushing the stale copy over newer server state.

### The fix

Updates omit those four fields and merge — which is exactly what the rule asks for: absent from the write, so the comparison is trivially satisfied, and preserved on the document because merge only replaces the keys it is given. Creating still writes the whole document, since `allow create` has no unchanged-fields test and a gallery must be born with the fields its schema requires. `flushUnsaved` replays in the update form for the same reason.

A unit test asserts both shapes and fails without the fix.

### Effect

- The three specs that flaked: **9 of 10 local runs clean**, from about 1 in 5 before, and each run drops from ~33s to ~16s because the 20s timeouts are gone.
- Full local chromium e2e: 266 passed, and the previously-failing `howto-form:70` passes.
- CI: three green shards, no retries.

### Corrections to my earlier comment

Two things I said above were wrong and are worth striking. The flake was **not** unexplained residue after the `array-contains` fix — it had a single specific cause, and I had not found it yet. And the `GalleryDatabase.edit` cache-ordering theory, which I described as a real deadlock, was not what was biting; I did not ship it, and the actual dirty-flag wedge is this permission denial instead.

PR #1380, opened 2026-09-12
Fix the WebKit nightly: a stalled Firestore stream, and the diagnostics that found it

amyjko:
# Context

The WebKit nightly failed **13 of the last 30 runs**, and **9 of the last 11**, while every Chromium run on `main` stayed green. The signal had become noise, so a genuine Safari regression would have been invisible in it.

It was never one regression. Runs [34600619076](https://github.com/wordplaydev/wordplay/actions/runs/34600619076) and [34692752819](https://github.com/wordplaydev/wordplay/actions/runs/34692752819) ran the *same commit* and disagreed.

## The cause

**Firestore's `experimentalAutoDetectLongPolling` probe intermittently leaves a WebKit stream that never delivers.** A `getDoc` then returns nothing at all — no error, no failed request, no `FirebaseError` — while the identical read completes in 2.9s on the next attempt.

Getting there took four hypotheses, and the diagnostics are what ended the guessing. The tests now print the page's visible text and its console events into the job log on failure, which produced this:

```
===== signed-out public space never loaded =====
url: http://127.0.0.1:5002/en-US/gallery/seed-public-gallery-00/howto
visible text:
🏠 Home / Galleries / Beginnings … —
events:
console[error] Couldn't get how-to with ID seed-public-howto-00: Error: read-timeout
```

Two measurements identified it:

- Raising `READ_TIMEOUT_MS` 8s → 20s changed **three** `read-timeout` lines per attempt into **exactly one**. The read wasn't slow; it never returned.
- A retry caught the *gallery* read stalling too, page frozen on "Loading, thanks for waiting!". Not a how-to problem — any read can hang.

Forcing long polling avoids the streaming transport entirely. **Scoped to the emulator**, because this file already documents why forcing it in production is harmful: it cycles many discrete requests instead of one stream and churns the session on large accounts with "Unknown SID" 400s. There is evidence of a stall against the emulator and none in production, so the fix goes where the evidence is.

## Result

[Run 34722454048](https://github.com/wordplaydev/wordplay/actions/runs/34722454048) on `6b6c7ec`, shard 2 — the shard that carried every cloud-read failure:

```
No files were found with the provided path: test-results/. No artifacts will be uploaded.
```

Playwright writes that directory only on a failure or a retry, so that is **zero failures and zero retries**, not merely a green tick. Its e2e step also fell to 12:00 from 14–16 minutes. **Zero `read-timeout` lines anywhere in the run.**

What remains is a different and much smaller class, worth their own issues rather than this PR: `announcer.spec.ts:301`/`:314` (the live region reads back `""` on first attempt — a timing race, no Firestore) and `key-pad.spec.ts:110`.

## Also here

Three of these are genuine fixes that turned out **not** to be what bit WebKit. They are kept because each is correct on its own, and the commit messages say plainly that they are not the cure.

- **A latched cloud read.** `resolveGallery` keeps an unreachable lookup as "still loading" on the promise that the caller re-runs it "whenever the maps change" — but a signed-out visitor has no listeners at all, so nothing ever did. One overrun read left a public space blank for the life of the page. `getHowTo`/`getHowTos` now separate "we never got an answer" from "there isn't one here for you", and the space re-asks on a backoff.
- **That backoff's cap**, 30s → 3s. Measured on Chromium: recovery after the network returns, 14.6s → 4.2s. Back off in proportion to what an attempt *costs* — an offline read is rejected in milliseconds, so doubling ran away in seconds and the page slept through a recovered network.
- **`READ_TIMEOUT_MS` 8s → 20s.** A budget, not a deadline: `raced()` abandons rather than cancels, so one set too low reports a read that would have succeeded as a failure. This matters beyond CI — a creator on a slow connection is the same case with no retry watching.

Two test races that only showed on a slower engine:

- **`offline-replay.spec.ts`** waited for the name to reach the *cloud*, then read it back from the *Dexie cache*. `waitForCachedProjectName` exists for exactly that precondition and says so in its docstring; the call site never used it.
- **`writing-vertical.spec.ts`** measured `.previews` through a locator, which re-resolves per call, so a re-render between the wait and the measurement reported a null width.

And CI hygiene: the workflow now names failed **and flaky** tests in the job summary (flaky was the invisible half — a shard going green on the second attempt is on its way to going red), and stops starting the Emulator UI, an extra Java and Node process per shard on a 3-core runner for a dashboard nobody opens.

## Related issues

- Related Issue #
- Closes #

## Verification

- `npm run check:now` — 3,506 files, **0 errors, 0 warnings**.
- `npm run test:run` — **532 files, 9,313 passed**, 7 skipped.
- New unit tests for the backoff policy; new e2e regression test for the recovery (cut Firestore before navigating, then assert the space fills in **with no reload**).
- Chromium e2e green as this PR's own gate.
- **WebKit**: `6b6c7ec` clean on the problem shard, with a confirmation run in flight. Scoreboard across every run tonight, including the failures:

| Head | Result | Cloud-read tests |
| --- | --- | --- |
| `b1daf5f` latch fix | green | both flaky |
| `b1daf5f` | red | `:151` failed all attempts |
| `ea81702` backoff cap | red | both failed all attempts |
| `a8a6e62` + diagnostics | green | both flaky |
| `c9c1386` 20s budget | red | `:151` failed all attempts |
| **`6b6c7ec` long polling** | **green** | **zero failures, zero retries** |

Accessibility: no user-visible UI, copy, color, or focus behavior changed. The only behavioral difference is that a how-to space that previously stayed blank forever now fills in on its own.

## Checklist

- [ ] Confirmation run(s) on `6b6c7ec`
- [ ] File follow-up issues for `announcer.spec.ts:301`/`:314` and `key-pad.spec.ts:110`
- [ ] `username.spec.ts:95` — the handle rename lands but the character rename sharing its **atomic batch** does not. Unexplained; did not recur on `6b6c7ec`, but that shard was clean throughout, so it may simply not have fired. Deliberately not guessed at.

Note for whoever touches this next: `importGraph.test.ts`'s landing-page budget sits at 4.1500MB against a 4.15 cap and counts raw source bytes including comments, which are stripped from the shipped bundle. It blocked a rationale comment three times here; each was absorbed by trimming rather than raising the cap, but it taxes exactly the comments CLAUDE.md asks for.


amyjko:
## Verification in progress — and a correction to this PR's diagnosis

The failing `webkit / test (webkit 2/3)` check is on **`b1daf5f`**, which is superseded. A fix aimed at exactly that failure is already pushed as **`ea81702`**, and a WebKit run against it ([34718618052](https://github.com/wordplaydev/wordplay/actions/runs/34718618052)) is in flight. No action needed on the red check itself.

### The control data is useful, so recording it

Two WebKit runs on `b1daf5f` (the 30s backoff cap):

| Run | Result | `seeded-load:151` |
| --- | --- | --- |
| [34716945641](https://github.com/wordplaydev/wordplay/actions/runs/34716945641) | green, 3 flaky | failed attempt 1, passed on retry |
| [34717934886](https://github.com/wordplaydev/wordplay/actions/runs/34717934886) | **red** | **failed all 3 attempts** |

That is the historical signature, so the first commit alone did not fix it.

### What the retry was actually doing

Instrumented locally against the emulator (Firestore cut for 20s, then restored, logging every request the page made):

```
 2.2s  REQ … Couldn't get gallery with ID seed-public-gallery-00
 2.7s  REQ …      ← 0.5s
 3.7s  REQ …      ← 1s
 5.7s  REQ …      ← 2s
 9.7s  REQ …      ← 4s
17.7s  REQ …      ← 8s
20.9s  restored firestore
33.8s  REQ …      ← 13 seconds of nothing
35.5s  RECOVERED
```

The retry fires exactly as designed. The page then **sleeps through a network that has already come back** — and that dead window, not the retry, is what blew the 30s budget.

**The premise in this PR's description is wrong.** It says a read hangs for `Database.READ_TIMEOUT_MS` (8s) before being abandoned, which would make doubling proportionate. It doesn't: once the client knows it is offline the SDK rejects in milliseconds without touching the network, so six failures pass in seventeen seconds and the delay is already sixteen. The latch this PR describes is real and retrying is the right cure; the timing story was not.

`ea81702` caps the backoff at 3s instead of 30s — *back off in proportion to what an attempt costs, not how many there have been*. Same probe, same cut:

| Cap | Retry gaps | Recovery after the cloud returns |
| --- | --- | --- |
| 30s | 0.5, 1, 2, 4, 8, 16 | **14.6s** |
| 3s | 0.5, 1, 2, 3, 3, 3 | **4.2s** |

~1.8s of the remainder is the Firestore SDK's own reconnect, which is not ours to shorten.

### Still open

- **Not yet proven on WebKit.** Only runs on `ea81702` count; I'll post the tally once there are two or three.
- `username.spec.ts:95` — the handle rename lands but the character rename sharing its *atomic batch* does not. That shouldn't be possible, and the artifacts that would settle it are on a host unreachable from where I'm working, so it stays unexplained rather than guessed at.

`npm run check:now` and `npm run test:run` are clean on `ea81702` (0 errors, 0 warnings; 532 files, 9,313 tests).

---
_Generated by [Claude Code](https://claude.ai/code)_

amyjko:
## The diagnostics answered it, and it wasn't the latch

The WebKit failures now print their own evidence into the job log. From [run 34719747264](https://github.com/wordplaydev/wordplay/actions/runs/34719747264), shard 2:

```
===== signed-out public space never loaded =====
url: http://127.0.0.1:5002/en-US/gallery/seed-public-gallery-00/howto
visible text:
🏠 Home / Galleries / Beginnings
How-tos
How-tos are short explanations and example code for how to do something with Wordplay…
—
events (last 4):
console[info] Successfully preconnected to https://firestore.googleapis.com/
console[error] Couldn't get how-to with ID seed-public-howto-00: Error: read-timeout
console[error] Couldn't get how-to with ID seed-public-howto-00: Error: read-timeout
console[error] Couldn't get how-to with ID seed-public-howto-00: Error: read-timeout
===== end =====
```

Three things fall out of that.

**The gallery resolves.** Not one `Couldn't get gallery with ID` line, and the page rendered its breadcrumb, the gallery name, the blurb and the `—` empty marker. So the latch this PR opened with is real but is *not* what fails on WebKit.

**The how-to read is killed by our own budget.** `read-timeout` is `Database.read` giving up at 8s. No `unavailable`, no `permission-denied`, no failed request — the page preconnected fine. Three reads × 8s ≈ 24s of the test's 30s, and the identical read finishes in **3.5s** once the page is warm.

**Retrying could never have helped.** `raced()` doesn't cancel the underlying `getDoc`, it stops waiting for it. Every retry re-races the same budget and abandons a read that was about to land — which is why capping the backoff at 3s (measured to take Chromium recovery from 14.6s to 4.2s) changed nothing here.

That makes this a **budget, not a deadline**, and a budget set too low reports a read that would have succeeded as a failure. `c9c1386` raises it to 20s. It matters beyond CI: a creator on a slow connection is the same case with no retry watching.

### Deliberately measured on its own

The other candidate is the transport — Firestore runs with `experimentalAutoDetectLongPolling`, and the comment in `firebase.ts` already anticipates falling back to `experimentalForceLongPolling` if detection misbehaves, which Safari has history with. Changing both at once would leave nobody able to say which worked:

- WebKit goes green → the read was merely slow, and 8s was starving it.
- WebKit stays red → the read never completes at all, no budget rescues it, and the transport is next **on evidence**.

### Scoreboard so far, unvarnished

| Head | Result | `seeded-load:151` / `:173` |
| --- | --- | --- |
| `b1daf5f` (latch fix) | green | both flaky |
| `b1daf5f` | **red** | `:151` failed all attempts |
| `ea81702` (3s cap) | **red** | both failed all attempts |
| `a8a6e62` (+ diagnostics) | green | both flaky |

Two red, two green — the flake never went away. The earlier commits fix a genuine defect (a signed-out visitor has no listeners, so nothing re-ran a failed lookup) and measurably improve Chromium recovery, but neither is what bites WebKit, and I'd rather the history say so than claim credit it hasn't earned.

`npm run check:now` and `npm run test:run` are clean on `c9c1386` (0 errors, 0 warnings; 532 files, 9,313 tests).

---
_Generated by [Claude Code](https://claude.ai/code)_

amyjko:
## It was the transport

`6b6c7ec` — forcing long polling against the emulator — produced the first clean result on the shard that carried every cloud-read failure. [Run 34722454048](https://github.com/wordplaydev/wordplay/actions/runs/34722454048), shard 2:

```
No files were found with the provided path: test-results/. No artifacts will be uploaded.
```

Playwright writes `test-results/` only on a failure or a retry, so that is **zero failures and zero retries** — `seeded-load:151`, `seeded-load:173` and `username:95` all live on that shard and none of them stumbled. Its e2e step fell to **12:00** from 14–16 minutes, and there are **zero `read-timeout` lines anywhere in the run**.

### Why measuring separately mattered

| Head | Change | WebKit |
| --- | --- | --- |
| `ea81702` | backoff cap 30s → 3s | red |
| `c9c1386` | read budget 8s → 20s | red |
| **`6b6c7ec`** | **force long polling (emulator)** | **clean** |

Had I changed the budget and the transport together, nobody could say which worked. The 20s run being red is what *proved* the budget wasn't the cause — and its failure shape is what identified the real one: three `read-timeout` lines per attempt at 8s became exactly one at 20s, so the read never returns rather than being slow.

### What is still flaky, and is a different problem

Not claiming these as fixed, and they predate this PR:

- `announcer.spec.ts:301` and `:314` (shard 1) — the live region reads back `""` on the first attempt. A timing race; no Firestore involvement.
- `key-pad.spec.ts:110` (shard 3) — "a key whose release never arrives stops repeating".

Both deserve their own issues rather than being folded in here.

### Still unexplained

`username.spec.ts:95` — the handle rename lands but the character rename sharing its **atomic batch** does not, which shouldn't be possible. It did not recur on `6b6c7ec`, but that shard was clean throughout, so this is not evidence that it is fixed. Left unexplained rather than guessed at.

Two confirmation runs are in flight; the flake rate has fooled me before and one clean run is one clean run.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #907, opened 2026-01-18
Automated tests for how-to permissions


melissachen2000:
### What are you trying to do that you can't?

See #825 for the feature description / design proposal and #882 for the original PR. Currently, there are no automated tests for this feature. 

### What is your idea?

See #882 for some manual tests that aim to verify core functionality (I currently cannot claim these are exhaustive). Furthermore, tests to verify permissions include the following:

# As a curator or creator in a gallery
- Can post a how-to
- Can post a how-to draft
- Can bookmark a how-to
- Can use reactions on a how-to
- Can start and add to a chat on a how-to
- Can delete a how-to or a draft
- Can edit a how-to or a draft
- Can see published how-tos in the Guide at `/guide`
- Can see published how-tos in the Guide panel in a project for the how-tos from that project's gallery
- Cannot see another creator's drafts in the Guide or in the how-to space
- Cannot edit someone else's how-to if they are not added as a collaborator 
- Can edit someone's else's how-to if they are added as a collaborator
- Can move any how-to

# As a curator of a gallery, in addition to the above:
- Can see the settings
- Can modify the settings (verify that new how-to prompt or reactions show up on how-tos created after the change)
- Can edit any how-to in the gallery, even ones where they are not specifically added as a collaborator

# As a viewer of a gallery (i.e., given expanded access)
- Creator Y hould have expanded access to gallery B if:
    - Curator X curates gallery A and gallery B
    - Creator Y is a creator or curator in gallery A 
- (If conditions are not met, then Creator Y cannot access the how-to space via the URL)
- Can see the gallery in the list of galleries with how-to access in `/projects` ~(NOTE: currently, this manual test does not pass, see TODO 1 below)~
- Can see the how-to space and navigate using the dropdown and center buttons, and go to a specific how-to using the URL
- Cannot add a new how-to
- Can bookmark, react, add project/how-to referenced, chat (the three social features on the right-hand panel of the viewer for a published how-to)
- When expanded access or public/private status is changed, no longer able to access the how-to space via the URL
- If added as a collaborator to a how-to, can edit or delete it
- Cannot move any how-to, unless added specifically as a collaborator

# As a user viewing a public how-to space
- Can see how-tos, if the gallery is public
- Cannot use any social interactions, including moving, commenting, reacting, etc. (the entire social pane should be removed)

# As a random other user (not added to the gallery as a curator / collaborator, not granted expanded scope access
- Cannot access the how-to page using the URL
[...]

PR #1384, opened 2026-09-12
Show a signed-out visitor a public gallery's how-tos (#1375)

### LINKED ISSUE #1375: A signed-out visitor sees no how-tos in a public gallery (webkit)
## What's the problem?

`tests/end2end/seeded-load.spec.ts:150` — "a signed-out visitor reads a public space and
cannot take part" — fails on webkit on `main`, reproducibly. Verified by stashing an
unrelated working tree, rebuilding, and running the spec alone against a fresh emulator:

```
✓ 4 [webkit] creator can open a seeded project (galleries/projects sync resolves)
✓ 5 [webkit] an expanded-scope viewer reads the space but cannot add to it
✘ 6 [webkit] a signed-out visitor reads a public space and cannot take part (30.4s)
```

It failed in 4 of 5 runs across two full `npm run end2end` passes and two `--last-failed`
reruns, so it is not purely timing.

```
Error: expect(locator).toBeAttached() failed
Locator: getByText('A how-to anyone can read').first()
```

The page itself renders — breadcrumbs, the "How-tos" heading, the space's introduction,
the bookmarks pane with "You must be logged in to use bookmarks". What is missing is the
how-to. The navigation combobox has only its empty option:

```yaml
- combobox "navigate to a how-to":
  - button "—"
  - option "—" [selected]
```

So the space loads for a signed-out visitor and its contents do not.

## What's the idea?

Worth deciding whether this is the product or the test. The rule's public branch is what
the test exists to cover — `isPublic` on the how-to, or `isGalleryPublic` on its gallery,
both of which `howtos`' read rule tests before any auth check — and `tests/rules/howToRules.test.ts`
passes, so the rule admits the read. That points at the client: which listener a
signed-out visitor gets for a public gallery's how-tos, and whether one is started at all.

Chromium passes, which is worth explaining rather than routing around.

## Who benefits?

Anyone following a link to a public gallery's how-tos without an account — the audience
the public branch exists for — and everyone reading CI, since a permanently red webkit
test trains people to ignore the suite.

Found while working on #8; unrelated to it.

amyjko:
# Context

`seeded-load.spec.ts`'s "a signed-out visitor reads a public space and cannot take part" fails on webkit on `main`. The issue asks whether this is the product or the test. **It is the product.**

A visitor's path to a how-to has **no listener and no retry**, so one slow read is the last word. Everyone else gets how-tos from a listener, and the page effect reads the database's `SvelteMap` synchronously — so it is *subscribed* to it, and a slow first read heals when a snapshot lands and the effect re-runs. **That reactive dependency is the retry.** A visitor starts no listener at all, because every query in `HowToDatabase` is built from a uid, and `getHowTo` swallows a failed read as `false`. Nothing re-ran, so a timed-out read left a public space permanently empty — the canvas and the navigation menu both render from that one list.

Two things rule out the obvious alternatives: rules admit the read (`firestore.rules` and `howToAccessScenarios.ts` assert `anon: ['read']`), and the navigation combobox passes a plain `label` with no `item` snippet, so the WebKit rich-option-snippet caveat doesn't apply — an empty combobox really does mean `howTos === []`.

## The fix

`getHowTos` retries what it couldn't reach, once, and answers `false` when it still can't. "We couldn't go look" is not "there are none" — the distinction `GalleryDatabase.find` already draws between `missing` and `unreachable`, reused rather than reinvented. Every caller was already written as `if (data)`, a guard that could never fail before and now means what it reads like; two of them even annotated a `false` they could never receive. The how-to space additionally guards on a generation, because these reads overlap when a gallery resolves more than once and a slow failing one could otherwise land last and empty a list that had arrived.

**Why this is WebKit's failure:** `playwright.config.ts` says it in the repo's own words — WebKit is ~2-3× slower than Chromium for Firestore round-trips. This route also still carries the language runtime (#1280) and builds a `Project` + `ConceptIndex` on mount, so the main thread is busy exactly while the 8s read budget runs.

Not taken: a gallery-scoped listener not gated on `$user`. It is the structural answer and it is bigger, and it has a trap worth recording — `handleSnapshot`'s cache GC prunes against `listenerDocIds` once it reaches `expectedHowToListeners`, which is **0** for a visitor, so an uncounted listener that registers a key trips the GC immediately. It would also want an anon `list` rules test; `howToRules.test.ts` only ever exercises `doc.get()`.

## Related issues

- Closes #1375
- Related Issue #907, #1351, #1280

## Verification

- `npx playwright test seeded-load --project=webkit` against a seeded emulator — 6/6 pass, including the failing test.
- `npx playwright test howto gallery --project=chromium` — 11/11, covering the three other `getHowTos` callers this changes.
- `npm run check:now` clean; `npm run test:run` 548 files, 9561 passed; `npm run test:sweep` 2527 passed.

**I could not reproduce the flake locally** — the spec passes alone on `main`, which matches the report of failures only during full-suite runs, where four workers contend on the single-process emulator. So the retry policy is unit tested (7 cases, including that an absent how-to is never retried while an unreachable one is) rather than pinned to the flake. Note that webkit runs **nightly only**, so this PR's own CI will not exercise it.

The helper is a leaf of its own because folding it into `HowToDatabase` saves nothing and that module can't be imported by a unit test on its own ("HowToDatabase is not a constructor"), which would leave the policy untestable at the level that reaches it. That costs +1 file on five import budgets, recorded there with the measurement (`projects` measured 4.2603MB against a 4.26 ceiling — a rounding boundary, not real growth).

**Accessibility:** no UI change; the same tiles render, they now arrive when the first read is slow. No new markup, colors, or focus targets, so screen-reader and keyboard passes are unchanged from `main` and I have not run them.

## Checklist

- [ ] Worth a reviewer's eye: whether the structural listener is wanted instead of, or in addition to, this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ These notes are publicly posted in [production](https://wordplay.dev/updates), s
 ### Fixed
 
 - 🎨 A heading inside a tabbed panel, like the ones in Settings, now sits closer to the part it names instead of drifting up toward the section above it.
+- 📖 A public gallery's how-tos now show up when you're not signed in. (#1375)
 
 ## 0.36.0 - 2026-09-09
 
--- a/src/db/howtos/HowToDatabase.svelte.ts
+++ b/src/db/howtos/HowToDatabase.svelte.ts
@@ -35,6 +35,7 @@ import {
 import { SvelteMap } from 'svelte/reactivity';
 import { v4 as uuidv4 } from 'uuid';
 import { z } from 'zod';
+import resolveHowTos from './resolveHowTos';
 
 ////////////////////////////////
 // SCHEMAS
@@ -855,6 +856,11 @@ export class HowToDatabase {
         return this.getHowTo(newHowTo.id);
     }
 
+    /**
+     * A how-to by id: `undefined` when it isn't there or isn't ours to see, and
+     * `false` when the read itself failed. The difference is the caller's to act
+     * on — only a failed read is worth asking again (#1375).
+     */
     async getHowTo(howToId: string): Promise<HowTo | undefined | false> {
         // do we have the how-to cached? return it.
         const howTo = this.howtos.get(howToId);
@@ -884,13 +890,21 @@ export class HowToDatabase {
         }
     }
 
-    async getHowTos(howToIds: string[]): Promise<HowTo[]> {
-        const results = await Promise.all(
-            howToIds.map((id) => this.getHowTo(id)),
+    /**
+     * The how-tos of a gallery, or `false` when we couldn't read any of them.
+     *
+     * Retries what it couldn't reach, because a signed-out visitor has no
+     * listener to heal a slow read: every query here is built from a uid, so
+     * their only path is this one, and a single timed-out read used to leave a
+     * public space looking empty (#1375). `false` rather than an empty list for
+     * the same reason — "we couldn't go look" is not "there are none", and every
+     * caller already guards on it before replacing what it is showing.
+     */
+    async getHowTos(howToIds: string[]): Promise<HowTo[] | false> {
+        const { howTos, unreachable } = await resolveHowTos(howToIds, (id) =>
+            this.getHowTo(id),
         );
-        return results.filter(
-            (ht): ht is HowTo => ht !== undefined && ht !== false,
-        );
+        return howTos.length === 0 && unreachable ? false : howTos;
     }
 
     ignore() {
--- a/src/db/howtos/resolveHowTos.test.ts
+++ b/src/db/howtos/resolveHowTos.test.ts
@@ -1,0 +1,86 @@
+import type HowTo from './HowToDatabase.svelte';
+import resolveHowTos from './resolveHowTos';
+import { expect, test, vi } from 'vitest';
+
+/** Only identity matters here, so stand-ins beat building real how-tos. */
+const one = { id: 'h1' } as unknown as HowTo;
+const two = { id: 'h2' } as unknown as HowTo;
+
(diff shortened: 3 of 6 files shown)


amyjko:
Superseded by #1380, which merged while this was being written, and which is the better fix.

We reached the same diagnosis independently — a signed-out visitor starts no listener, so the reactive map that heals a slow read for everyone else doesn't exist for them, and `getHowTo` swallowed the failure — and landed nearly the same shape: `getHowTos` returning `{ howTos, unreachable }`, `getHowTo` telling "couldn't go look" apart from "isn't there", and the page keeping what it has rather than blanking.

But #1380 goes further in two ways that matter:

- **It found the root cause.** Not a read that was merely slow: `experimentalAutoDetectLongPolling` intermittently leaves a WebKit stream that never delivers, so the `getDoc` returns nothing at all. Forcing long polling against the emulator removes the failure instead of recovering from it. I had gone as far as "8s budget plus contention" and stopped there.
- **Its retry is unbounded with a capped backoff** (`retryDelay.ts`, 500ms doubling to 3s), so a page left open across a sleeping laptop comes back on its own. Mine retried exactly once and then gave up, which would have left the same blank space for anyone whose outage lasted longer than 300ms.

Closing rather than rebasing: there is nothing here worth keeping on top of it.

One small thing that survives, if anyone wants it as a follow-up: the how-to fetch effect still has no generation guard, so two overlapping `getHowTos` can land out of order and an older list can overwrite a newer one. The sibling `resolveGallery` effect has had one for this reason. It is a much smaller problem than the one #1380 fixed — both results are real lists — so it may not be worth a PR.

#1375 itself looks fixed by #1380 and can probably be closed.

Issue #742, opened 2025-04-22
Custom characters not showing when not signed in


melaya206:
### What happened?

Custom characters show as a blank box in projects when opened in a browser not signed into Wordplay. Shows for any user, not just the creator.

**Screen signed in to the account that created custom emoji:**

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/c8762a95-16d0-4e7b-b4cf-357221be76dc" />

**Screen signed in to a different account:**

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/e7f94924-8571-4714-8a5b-8e5146adcf32" />

**Screen not signed in:**

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/dfdc3fc3-234c-4a01-8f22-c77c1edf78a2" />

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

macOS

amyjko:
This is the intended behavior. If the character is not made public, then it cannot be loaded and displayed. I'm relabeling this an improvement instead of a defect; can you think of better feedback to show instead of a box, that would make it clearer why it's not visible?

melaya206:
I think a disclaimer before playing the project would better communicate this, e.g., "This project contains private custom emojis, sign in to be able to see them", like with the moderation note.

Maybe a red "X" or "N/A" could also make this clearer.

amyjko:
This is a good start! Labeled this a starter project for someone who wants to do some basic design work.

Issues to consider:

* Where would this warning go? (The editor has a footer message area; maybe we add one to the stage too).
* What would the message be?
* Should the rectangle be a different shape, or is the message enough? Is there a shape that is clear enough that a message wouldn't be required?

PR #1390, opened 2026-09-13
Let a signed-out visitor watch a public gallery's how-tos (#1375)

### LINKED ISSUE #1375: A signed-out visitor sees no how-tos in a public gallery (webkit)

amyjko:
# Context

A signed-out visitor had no realtime listener of any kind. Every how-to listener is uid-scoped and torn down when there is no user, so nothing ever re-ran a visitor's lookups: one read that overran the 20s budget left a public space blank for the life of the page, with no error and no way back but a reload. That is the structure behind #1375's WebKit failure, which #1380 treated at the symptom by adding a backoff.

Galleries and characters both already have a public, non-user-scoped read path — #742's lesson, "deliberately not gated on being signed in: the security rules have always allowed anyone to read a public character." How-tos had the one-shot `getHowTo` but no public listener, while `howToAccessScenarios.ts` has recorded `anon: ['read']` all along. The server always said yes; the client never asked twice.

`Galleries.watchPublic(id)` is what a surface calls now. It holds a listener on the gallery document and, through `HowTos.watchGallery(id)`, a `galleryId == X && published == true` query over its how-tos — both ref-counted through a new `src/db/Watchers.ts`, so the four surfaces that show a gallery's how-tos share one subscription.

**It costs fewer reads, not more.** `watchGallery` picks exactly one path rather than watching *and* reading through:

| Viewer, cold load | Before | After |
| --- | --- | --- |
| Signed out, public gallery | N (up to 3N across surfaces) | **N**, one query |
| Signed-in member, own gallery | up to 3N, plus the listener snapshot | **0** extra |
| Member with notifications off | up to 3N | N + drafts |
| Moderator, private gallery | up to 3N | ≤ N, deduped |

`getHowTo` also shares a read already in flight, which is the other half of that — there was no pending map, so surfaces mounting together on a cold cache each issued their own `getDoc` for the same id.

Three things are load-bearing and easy to get wrong, each with a regression test:

- **The query must not filter on `isPublic`.** A how-to in a public gallery is readable through the rule's `isGalleryPublic` branch whatever its own flag says, and most are `isPublic: false` — including the seed fixture. That filter matches nothing and reads as a broken listener.
- **Expanded scope is checked *after* the public branch.** Listener 3 carries `scopeOverwrite == false`, so on a gallery that is also public it would silently omit a how-to the rules admit.
- **`startListening` has to re-decide the watches.** `listen` defers to idle, so a gallery the uid listeners now cover would otherwise keep the public query it started while they were down, and pay twice.

The cache GC needed rework to survive a listener that starts and stops on navigation — a count breaks in both directions, so it is now a set of expected keys plus a `directReadIDs` exemption keeping the invariant that *collection may only evict what a listener put in the cache*. That independently closes a latent bug: a signed-in non-member's how-tos from someone else's public gallery were evicted by the next snapshot from any of their own listeners.

Two smaller fixes a visitor needed: the how-to page no longer gates its concept index on being signed in, and `HowToConceptView` no longer bumps a view count it cannot write (a guaranteed `permission-denied` per how-to rendered). The same-looking ternary in `Guide.svelte` was never a gate — `getUser()` returns the store, so it was always the truthy branch — but it goes too, so nobody "fixes" it the other way.

**+1 file on every page graph** (`Watchers.ts`, no imports of its own) and a hundredth on each byte budget, both recorded in `importGraph.test.ts`.

## Related issues

- Closes #1375

## Verification

`npm run check:now` clean. 9,636 unit tests pass, including new coverage for the query shape (asserting the *absence* of `isPublic`/`scopeOverwrite` constraints), subscription sharing, the sign-in transition, `ignore` vs `stop`, the GC exemptions, and that a signed-out visitor's watch issues **zero** `getDoc`s — the regression test for the double-read mistake.

**Rules: 97 tests, six of them new and query-shaped.** No rules test in the repository drove a query before — they all used `doc.get()`, which proves nothing about a listener, since Firestore evaluates a query against every document it matched and rejects the whole thing if one is denied. The new block puts a draft beside the published how-to so "all or nothing" is real.

**Browsers.** Chromium: `seeded-load` 7/7 and `accessibility` 31/31. **WebKit — where #1375 was reported and which does not run on PRs — `seeded-load` + `accessibility`, 38/38**, including "a signed-out visitor reads a public space" and the cut-Firestore/restore test that recovers with no reload. That last one is what says dropping the gallery-branch retry was safe.

The full local e2e run showed 8 WebKit failures, all investigated rather than assumed:
- `chrome-selection` (2) fails on stashed clean `main` too — pre-existing.
- `gallery-sharing` (2), `kit-editing`, `optional-field-persistence`, `howto-form`, and `accessibility-authed`'s characters scan all pass in isolation with this branch — load flakes under four workers against the single-process emulator.

**Accessibility.** `/gallery/seed-public-gallery-00/howto` gained a `ContentMarkers` entry, without which axe scanned before the subscription delivered and the pass meant nothing. Both color schemes pass on Chromium and WebKit. No new colors, no new interactive elements — the change is which data reaches surfaces that already existed.

## Checklist

- [x] Screen reader pass (VoiceOver) — not done; no new UI, but a how-to space now populates for a signed-out reader, so the canvas is worth one listen.
- [x] Manual two-window check: open a public gallery's how-to space signed out, have a curator rename the gallery and publish a new how-to, confirm both arrive with no reload. This is the actual promise and no automated test asserts it.


amyjko:
Two CI cycles failed before this was right, and the first diagnosis was wrong — recording what it actually was.

**The cause was a feedback loop in the watch effect, not a wedged stream.** Acquiring the watch reads the gallery maps to decide what to subscribe to, and the watch then writes each snapshot back into `publicGalleries`. Read inside a Svelte effect, those reads become the effect's own dependencies: it tore its subscription down and started another on every snapshot, and never lived long enough to deliver one. Silently — which is why the CI dump showed a resolved gallery, no how-tos, and `events (last 0)`.

Measured against the built app with the emulator seeded, counting Firestore listen requests over twelve seconds on the how-to space:

| | listen requests |
| --- | --- |
| a page with no watch | 5 |
| before | 5139 |
| after | 10 |

The fix is `untrack` around acquiring the watch, plus keying the effect on the gallery **id** rather than the gallery object, which is rebuilt on every snapshot even when nothing about it changed.

One more real bug found on the way: a read-through for a gallery held in none of the local maps reported itself unanswered and retried forever. There is no list of how-tos to ask for in that state, so it is done, and whatever makes the gallery known re-decides the watch.

**On the per-document fallback in `a9299e6e5`:** its commit message blames a wedged stream, and that was not what was happening here. I have kept it rather than reverted — a stream that wedges after a transport interruption is a real failure #1380 documented, and the net costs nothing while the watch is working — but it is not what fixed this, and the later commit says so.

Issue #564, opened 2024-09-12
Allow installation as progressive web application


amyjko:
### What are you trying to do that you can't?

The application is fundamentally a web app currently. This is in tension with many teacher use cases, which involve installing an application, using it offline, and storing projects locally on a device instead of in the cloud, and allowing it to appear as a native application on a device rather than requiring navigation to a website.

### What is your idea?

Support Wordplay as a "Progressive Web App". There are a variety of tradeoffs to this model, but it is increasingly well supported. There's much to decide about Android, iOS, and macOS integration, so the design needed here is to evaluate all of the possible tweaks that we would make to support this use case and develop a design proposal that provides a concrete list of modifications and additions to make this possible. There's a good overview here of the [iOS PWA limitations](https://brainhub.eu/library/pwa-on-ios). Other tradeoff analyses likely exist.

jsorror:
@amyjko can I be added to this issue? Thank you!

diana24p:
@amyjko I would like to work on this issue can I please be assigned

vedashnii:
@amyjko I'd like to work on this issue, could I be assigned to it? Thank you!

anushkamalpani:
@amyjko Can I be added to the issue?

amyjko:
Are all four of you willing to work together? If you're not sure, create a channel in Discord for this issue (per the onboarding guide) and discuss there, and post a decision here.

jsorror:
[splash_screens.zip](https://github.com/user-attachments/files/17404698/splash_screens.zip)

vedashnii:
[splash_screens.zip](https://github.com/user-attachments/files/17404702/splash_screens.zip)
Light Mode Splash Screens + Metadata

jsorror:
Status bar fix to have transparent background:

In metadata, add:

`<meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">`

jsorror:
The IOS app still has text selection, highlighting, and callouts enabled. To disable these:

To disable selection, set the `-webkit-user-select` CSS property to `none`

`body {
   -webkit-user-select: none;
}`


To disable highlighting, set the `-webkit-tap-highlight-color` to `transparent`

`body {
   -webkit-tap-highlight-color: transparent;
}`


To disable callouts, set the `-webkit-touch-callout` to `none`

`body {
   -webkit-touch-callout: none;
}`

Vraj1234:
Hi guys, 

[...]

PR #1410, opened 2026-09-16
Serve the pages we prerender

### LINKED ISSUE #1409: No route but / serves its prerendered page, so per-locale metadata never reaches a reader
## What happens

No route except `/` serves its prerendered page. Everything else gets `200.html`, the 60 KB SPA shell. Measured against production:

| URL | serves | bytes |
| --- | --- | --- |
| `/` | prerendered `index.html` | 415,609 |
| `/about` | **shell** | 60,545 |
| `/about.html` | prerendered | 379,340 |
| `/guide` | **shell** | 60,545 |
| `/guide.html` | prerendered | 538,901 |
| `/es-MX` | **shell** | 60,545 |
| `/es-MX.html` | prerendered | 415,858 |

The prerendered files are deployed and reachable, but only by an explicit `.html` URL that nothing links to. `firebase.json` sets no `cleanUrls`, so `/about` matches no file and falls through the `**` → `/200.html` rewrite.

**That is 44 files and 16.8 MB of HTML built on every deploy and never served.**

## Why it matters

**Per-locale metadata never reaches a locale route.** `hooks.server.ts` injects the manifest path, `og:image:alt`, the `noscript` text and the unsupported-browser copy per locale — and only `/` (which is en-US) ever receives them. In production:

| URL | manifest served |
| --- | --- |
| `/` | `/manifests/en-US.webmanifest` |
| `/es-MX` | `/manifests/en-US.webmanifest` |
| `/ja-JP` | `/manifests/en-US.webmanifest` |

So someone installing Wordplay from a Spanish URL gets an app named in English — which is the thing per-locale manifests were added for (#564).

**Social previews and crawlers** get the shell for every route but `/`: no title, no description, no content.

## The catch, which is why this needs a decision rather than a patch

`"cleanUrls": true` in `firebase.json` fixes it in one line — verified against the hosting emulator: `/es-MX` then serves the prerendered file, carrying `/manifests/es-MX.webmanifest`.

But it is not obviously the right fix:

- **Cost**: +16–18 KB brotli per route (shell 13.2 KB → prerendered 29–32 KB), on HTML that is not immutably cached.
- **The prerendered body is always English.** SSR has no `navigator`, so every per-locale page is prerendered in English. Serving it to a first-time non-English reader paints English content and then swaps it — a flash of English where the shell currently shows nothing. `locale-loading` only hides that for readers whose *stored* preference is non-English, which a first-time visitor does not have.

So the real question is which of these is wanted:

1. **Serve them** — fix the metadata and give crawlers content, accept the bytes, and deal with the English flash (possibly by having `hooks.server.ts` do more, or by extending `locale-loading`).
2. **Stop generating them** — if the app is an SPA shell in practice, `entries()` in `src/routes/[[locale]]/+page.ts` is producing 31 locale copies for nothing, and the same goes for the other prerendered routes.

Either way the current state is the worst of both: the build pays for the pages and no reader gets them.

## Related

Found while measuring locale load timing for #1408, which deliberately puts its `<link rel="preload">` in the shell *because* that is what a locale route actually serves.

amyjko:
# Context

Closes #1409.

`getSitemap` advertises **337 URLs**. Every one except `https://wordplay.dev` served `200.html` — the SPA shell, which has **no `<title>` at all** and 385 characters of body text whose entire content is the noscript message, *"Wordplay uses JavaScript."* A crawler or unfurler that doesn't run JavaScript saw that for `/about`, `/guide`, `/learn`, every locale route, and every gallery and project link.

The pages existed the whole time. `firebase.json` set no `cleanUrls`, so `/about` matched no file and fell through the `**` rewrite; only `/about.html` reached it, and nothing links that way. **355 prerendered HTML files were built on every deploy and never served.**

## The change

`cleanUrls: true` is the whole fix. Two things had to go with it.

**The flash.** The prerendered body is always English — SSR has no `navigator` — so serving it to a first-time reader at `/es-MX` would paint English and swap it. `locale-loading` now also applies when the URL names the locale, not only when the stored setting does. That only became reasonable after #1408: before it, the locale wasn't requested until the JS bundle had run, so hiding meant a blank page for ~300 ms.

**The reveal.** `localesReady` reports the locales the *constructor* asked for — the browser's and the stored setting's. For a first-time reader with an English browser that's en-US, which is bundled, so it's true immediately and the reveal fired before Spanish arrived. The layout now also waits for the URL's locale to be the one in effect.

## Verification

Hosting emulator, first-time reader (empty storage) at `/es-MX`:

- Renders in Spanish with **no flash of English** — `locale-loading` applied at 89 ms, removed at 492 ms.
- `/about` → 381 KB, `/es-MX` → 418 KB, `/es-MX/about` → 381 KB (per-locale sub-routes are prerendered too). `/about.html` 301-redirects to `/about`.
- `cleanUrls` does **not** shadow the `getPagePreview` rewrite — `/project/…` and `/gallery/…` return identical bytes with and without it. Checked against a control run, because the emulator doesn't route those to the function either way, which would otherwise have read as a regression caused by this change.
- `npm run check:now` clean; unit suite green; **63 e2e tests** pass (`accessibility`, `localize-guidance`, `localize-glossary-forms`, `writing-vertical` — the specs that navigate locale-prefixed routes).

## What this does not do

**It is not faster.** Time until Spanish text is visible, three runs each:

| | runs (ms) | mean |
| --- | --- | --- |
| With this change | 425 / 442 / 439 | 435 |
| Without | 473 / 428 / 438 | 446 |

That's noise. An early single run suggested 343 against 482 and I nearly reported it; repeating it killed the claim.

So the ledger is honest: **336 URLs stop serving an empty titleless page**, at a cost of **+16–18 KB brotli per route** (shell 13.2 KB → prerendered 29–32 KB) on HTML that is not immutably cached. It buys nothing a reader can see — the value is entirely for crawlers and social unfurlers.

A correction to #1409 while I'm here: **I claimed the per-locale manifest was broken and it isn't.** `+layout.svelte` already rewrites the manifest link once the locale resolves, and a real browser on production confirms `/es-MX` gets `/manifests/es-MX.webmanifest` and `lang="es"`. My original table was what `curl` sees — the pre-hydration document — and I drew the wrong conclusion. #564 was never regressed.

## Checklist

- [x] **Confirm on `test.wordplay.dev` that `/project/…` and `/gallery/…` still reach `getPagePreview`.** The emulator doesn't exercise function rewrites, so that routing is genuinely unverified here, and it's the one thing `cleanUrls` could plausibly break.
- [x] Worth a look at whether the +16–18 KB per route is a trade you want, given it buys no reader-visible benefit.


amyjko:
## Verified on `test.wordplay.dev`

Deployed this branch to dev (`hosting` scope) and measured.

**The risk is cleared.** All three function rewrites still fire:

| URL | routed to |
| --- | --- |
| `/project/…` | `getPagePreview` ✓ |
| `/gallery/…` | `getPagePreview` ✓ |
| `/sitemap.xml` | `getSitemap` ✓ |

Detected by `x-cloud-trace-context`, which function responses carry and static files don't — positive evidence the function ran, rather than an unchanged byte count that a fallback could also explain.

**Prerendered pages are served**: `/about` 60,545 → 381,298 B, `/es-MX` → 417,816 B, `/es-MX/about` → 381,512 B. Per-locale manifests now arrive **without JavaScript** (`/es-MX` → `/manifests/es-MX.webmanifest`, `/ja-JP` → `/manifests/ja-JP.webmanifest`).

**Crawler-visible body text on `/about`: 385 → 37,301 characters.**

## Correction to this PR's description

I wrote that "336 URLs stop serving an empty titleless page." That is **half right**: they stop being empty, and they are **still titleless**.

`/about` and `/guide` serve no `<title>` at all; `/` and `/es-MX` serve a generic `<title>Wordplay</title>`.

This is by design and out of reach here. [Title.svelte](https://github.com/wordplaydev/wordplay/blob/main/src/components/widgets/Title.svelte) sets `document.title` **imperatively rather than emitting `<title>` inside `<svelte:head>`**, and says why: Svelte 5 inserts hydration anchor comments alongside reactive text, and `<title>` is parsed as RCDATA, so those comments would survive as literal characters in the title. Being imperative, it only runs in a browser, so prerendering can never capture it.

So the accurate ledger for a crawler that does **not** run JavaScript:

- **before**: no title, 385 characters of "Wordplay uses JavaScript"
- **after**: no title, 37,301 characters of the actual page

Better, and not the whole fix. A per-page title for non-JS crawlers needs the `<svelte:head>` problem solved separately, and is worth its own issue rather than being smuggled in here.

Issue #1062, opened 2026-04-26
Learnability of markup syntax for highlight


melissachen2000:
### What are you trying to do that you can't?

We added a highlight annotation for Examples to the markup language in #1061 , but it is currently not very learnable outside of the tooltip and an one-line mention in the guide.

### What is your idea?

None at the moment. It might potentially be interesting to have some part of the tutorial dedicated to markup, or a separate tutorial?

PR #1418, opened 2026-09-19
Let a program's data be saved as a CSV or JSON file

### LINKED ISSUE #157: Export table as CSV
## What's the problem?

Getting data out of Wordplay is important. It should be possible to select a table literal node or a table value and copy it to the clipboard as a CSV for export into other applications.

## Who benefits?

Anyone who wants to use Wordplay to do data processing.

## Design ideas

* Update copy command to copy table as CSV mime type
* Add a copy to clipboard button to the table value view

amyjko:
# Context

Wordplay could read data in but never write it out: pasting CSV into the editor already becomes a `TableLiteral` (`interpret.ts`), and there was no inverse. A student who builds a table of survey results could only look at it.

A program's data — a table, a list of records, a map, a structure — can now be saved as a spreadsheet (`.csv`) or a JSON file, from two places:

- **On stage**, under the value, when a program's output is data rather than a `Stage`.
- **In the editor**, as the last item of the node menu on a selected node (`⌃↓`, or the `▾` handle in blocks mode), plus a `⌃⇧E` chord.

Both open one dialog that shows the actual bytes, with **Save** and **Copy**. Copy matters as much as Save: pasted back into the editor, an exported CSV rebuilds the table, so the feature closes a loop rather than opening an exit.

## Related issues

- Closes #157

The issue proposed a CSV mime type on the copy command plus a button on the table value view. This does the second and replaces the first with a dialog, because the format is a choice (CSV or JSON) and because seeing the file before writing it is most of the value for a learner. `#157` is labelled `needs design`; the design decisions are below.

## What decides where the affordance appears

Three constraints, and together they settle it: it must be **discoverable**, it must **not appear when inactive**, and it must **cost nothing on the editing path**.

- **`canExport` is O(1)** — a class test plus, for a collection, its first element. It is the only thing a render path, a menu build or a `Command.active` may call. `toGrid`/`toJSON`/`serialize` walk the whole value and run **only inside the open dialog**, which [`exportImportConvention.test.ts`](src/values/export/exportImportConvention.test.ts) enforces. The two can disagree — a list whose hundredth item is a function passes the cheap test — and the dialog says so rather than walking a long list on every frame that renders it.
- **Not in the editor toolbar.** `OverflowToolbar` rebuilds its `ResizeObserver` set whenever its item count changes, so an item that hides when inactive thrashes layout as the caret crosses boundaries — the documented reason the markup toolbar greys its annotations instead (#1062). Greying an item that is inactive almost everywhere was the other option. The menu costs nothing because it is built when it opens.
- **Not on the debug value bubble.** In debug mode every rendered expression has one and they re-render per edit.
- **A lone scalar is not exportable.** A number or a text value is not worth a file with these two formats, and it is also what keeps the item off every number literal's menu.
- **The menu item is last.** Changing the program is what that menu is for. It also keeps the opening selection (`[0, undefined]`) on the first suggestion rather than preselecting the action.

## A bounded widening of the menu

`Menu.revisions` is untouched: actions ride in a field of their own and are concatenated onto the end of the organization, so the purpose grouping, `soundRevisions` (a parser round-trip that means nothing for an action) and `removeDuplicates` still only ever see a `Revision`. `getActionsAt` is a new function beside `getEditsAt`. That leaves three one-branch edits — `Menu.doEdit`, the type-ahead in `Menu.svelte`, and one new sibling component, `MenuActionItem.svelte`, rather than widening `MenuItem`, every line of which is about previewing the code a revision would produce.

## What the formats do to a value

- **A number exports as a number.** A column whose numbers share one unit carries it in the header (`rain (mm)`), so a spreadsheet can total the column; a mixed-unit column keeps them in the cells. `TableLiteral.from` keeps the first `Name` token of a header cell, so the annotation survives re-import as the bare name.
- **Booleans are `true`/`false`, not `⊤`/`⊥`**, because that is what `TableLiteral.from` reads back.
- **JSON is rendered by hand**, not through `JSON.stringify`, so a number keeps `Decimal`'s exact text — Wordplay uses decimal.js precisely because binary floats lie. Units and non-finite numbers can't be expressed, so they're dropped and the dialog says which.
- **The download carries a UTF-8 BOM and the clipboard does not.** Excel misreads unmarked UTF-8 and Wordplay content is very often not Latin; a BOM pasted back would land inside the first header cell.

## Three bugs in the round trip

The invariant is that a table exported as CSV and pasted back is the same table ([`roundTrip.test.ts`](src/values/export/roundTrip.test.ts)). Three existing defects were in the way, all of which corrupt the *existing* paste-CSV path too:

1. `TableLiteral.from` read `'false'` as `BooleanLiteral.make(true)` — silently wrong rather than merely broken.
2. `isCSV`'s character class was `[a-zA-Z0-9.%&-() _]*`, so CSV containing non-Latin text or emoji never matched at all.
3. A text cell was rebuilt by rejoining its *tokens* with spaces, so `likes, commas` came back as `likes , commas`; and quote-stripping took a quote off each end independently, so `and "quotes"` lost one.

Each fix was confirmed by reverting it and watching its own test fail.

## Verification

**Automated**

- `npm run check:now` — clean.
- `npm run test:run` — 614 files / 10,465 tests.
- `npm run test:sweep` — 2,594.
- `npm run end2end` — **chromium fully green** (581 passed, zero chromium failures). The webkit failures the local runs show are emulator contention, not regression: the failing set differs run to run, and the same specs in the same isolation give 28/0 on `main` and 29/0 here (the extra is the new export test). The run before the locale fix is what surfaced the crash above.
- New unit coverage: CSV escaping and the reader/writer round trip, the grid decider over every value shape, the unit-in-header rule, JSON decimal fidelity, the reachability rule, `canExport` against `formatsFor`, and the import convention test (verified to fail when violated).
- One e2e assertion added to an existing spec rather than a new test: the dialog passes axe in both schemes, switching format re-serializes and renames the file, and **Save** produces a real download named `Pet Survey-Table.csv`.

**Keyboard** — `⌃⇧E` is free of any other command and absent from `BrowserReserved`, and no command may declare Control+Alt. The menu path (`⌃A` to select, `⌃↓` to open) was verified to reach the item and open the dialog; the dialog is the standard `Dialog`, so focus moves in on open and Escape closes. Verified in a browser: Enter on the menu item opens the dialog, focus lands inside it, Tab cycles close → format → file name → Save → Copy and wraps without escaping, Escape closes, and `⌃⇧E` opens it from a selection. (The JSON radio is not its own tab stop, which is correct for a radiogroup — arrows move within it.)

**Color** — the preview panel initially used `--wordplay-inactive-color` as a background, which is `--color-grey-text`, a *text* token: black on it measures **3.94:1 in light and 3.28:1 in dark**, both under AA, and axe did not catch it. It now uses `--wordplay-alternating-color`, the surface the AA text variants are guaranteed against: **18.9:1 and 16.7:1**. Nothing conveys meaning by color alone. `MenuActionItem` declares `--wordplay-hover-text` on the gold hover ground, as `hoverContrast.test.ts` requires.

**Screen reader** — **not yet done, and it needs a pass.** The announcements are wired (`export` kind, queued lane) and each names the file or the format so a second save in a row isn't dropped as a duplicate; the command declares `feedback: 'focus'`; the `<pre>` deliberately carries no `aria-label` so its content stays navigable. But no VoiceOver or NVDA pass has been made against the dialog or the new menu item.

## Localization, and a crash it caught

Sixteen new `[plain]` strings plus one `[formatted]` explanation under `ui.export`.

Landing them in en-US alone **broke the app in every other language**, and the full e2e run is what caught it: sixteen `writing-vertical` and `validation-rtl` tests failed with the page stuck on its loading screen. `Locales.get` runs the accessor against each chosen locale and falls back to en-US only when the result is `undefined` — a *missing key path* throws on the way there (`l.ui.export.label` where `l.ui.export` is undefined). A locale that hasn't been padded with a new key is a broken app in that language, not merely an untranslated one.

`npm run locales-fix` pads all 30 with `$?`-marked English, which is what this PR carries. The strings are **not translated** — that is a deliberate separate pass — but the app now falls back to English in every locale instead of dying.

The invariant is now written down in `CLAUDE.md`: adding a string to `UITexts.ts` is a two-step change, and only the non-English e2e specs catch the second step being skipped.

**CI's `unit / locales` job will be red**, with exactly 30 findings — “18 unwritten strings would fall back to English” per locale — and nothing else. That is the documented state for new content before its translation pass.

## Import budget

`galleries` moves by 0.01MB for the new en-US strings, which every page carries. **No file count moves**, and nothing that serializes a value is on any page graph: `ValueExportButton` loads the dialog through a dynamic `import()`, because a static one put that page over its budget — recorded in `importGraph.test.ts` so the door stays visible.

## Checklist

- [ ] VoiceOver pass over the dialog and the menu item
- [ ] Translate `ui.export` into the other 30 locales
- [ ] Decide whether the class-roster CSV (`teach/class/new/+page.svelte`) should move onto the new writer — it escapes nothing, declares `charset=utf-16` while writing UTF-8, and leaks its object URL


amyjko:
Followed up by putting the class roster page on the same reader and writer rather than leaving a second copy of CSV in the tree.

It had its own on **both** sides, and both were wrong:

- **Reading** split on commas by hand, so a spreadsheet's `"O'Brien, Mary"` became three cells. In an email class that puts `"O'Brien` in the address column, `everyRowHasAnAddress` rejects it, and the teacher simply cannot create the class from a roster containing a comma in a name.
- **Writing** joined strings with no escaping, declared `charset=utf-16` while a `Blob` writes UTF-8, and never revoked its object URL.

Reading now goes through `parseCSV` and moved into `roster.ts`, beside the rest of what a row means — that file already existed to be the pure, millisecond-testable home for exactly this, and the parsing had no business being a `$derived` in the component. Five new cases cover quoted separators, doubled quotes, blank lines and non-Latin cells.

Writing goes through `writeCSVRows` + `downloadBytes`. The byte order mark that lets Excel read a non-Latin file is now one rule in `csvFileBytes`, which the value export calls too instead of keeping its own copy — so the "download gets a mark, clipboard does not" decision exists in exactly one place.

